PR #3, opened 2016-07-20
remove repetitive logging in db.js

evanshortiss:


fh-automatron:

evanshortiss:
We're seeing some logs too frequently in our application, and there's no way to disable them so this PR removes them as a quick patch. 

Of course the SDK should be possible to debug with some logs, but they should be configurable and have logging, just not by default for customer as it clutters our log files 👍 

Some options:

```js
var fh = require(fh-mbaas-api)({
  // Defaults to "error" or similar
  logLevel: 'info'
});
```

```js
var fh = require(fh-mbaas-api);
// It should be disabled by default
fh.enableLogging();
```

A third option might be to use the [debug](https://www.npmjs.com/package/debug) module, or [fh-bunyan](https://www.npmjs.com/package/fh-bunyan) which can be configured somewhat based on FH_ENV.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fh-mbaas-api",
-  "version": "5.13.1",
+  "version": "5.13.2",
```
matzew: Perhaps it's better to increase only before a real release?

lib/db.js
```diff
@@ -122,7 +119,6 @@ var db = function (params, callback) {
 
     var fhdb = require('fh-db');
     if (process.env.FH_USE_LOCAL_DB || process.env.FH_MONGODB_CONN_URL) {
-      logger.debug('using local_db');
       return fhdb.local_db(params, callback);
     }
     else {
```
matzew: 👍

package.json
evanshortiss: Sure, will undo that. I'm used to being told to do it in other PRs!

shannon:
@david-martin It looks like you added this logging. 

https://github.com/fheng/fh-mbaas-api/commit/4e06816032de1a791d8edfea18d0dac43fab29e8

Do you have any insight as to why this is would be outputting after every db call? Assuming this logging was added for a reason, does it make sense to remove it until a more configurable logger is implemented?

package.json
matzew: I think we should establish a more reasonable way to release modules, and to version bumps at release time /cc @grdryn

grdryn: Yes, I'm all for improvements like that. What is a real release? Can this change be released? It wouldn't be included as the default in a version of the platform or templates until the next platform release, but maybe it's useful to @evanshortiss before that?

evanshortiss: @grdryn absolutely, we'd appreciate a patch release whenever it's feasible for you folks 👍  We're not too concerned about templates here 😄

david-martin:
@shannon Rather than removing logging, I would expect to be able to change the level to include this extra output. If the default loglevel is outputting 'debug' output, then this should probably change.
So probably 2 suggestions here:
- change the default loglevel to something more appropriate for production e.g. 'info'
- allow changing the loglevel

evanshortiss:
@david-martin that would be ideal, but at present a magic number configures this before we can possibly change it from our application layer, and even at debug level only 1/4 of the logs would be removed. 

Ultimately it needs more work, but would be great to get rid of the current behaviour while eng can decide on an alternative SDK/logger structure.

```
var logger = new consolelogger.ConsoleLogger(3);
```

david-martin:
Thanks for the info @evanshortiss. Any suggestions on how to improve the sdk are appreciated. 
@matzew should we get a jira issue in the public FH namespace on the go? I think it would be good to get community input on this.

evanshortiss:
@david-martin @matzew that would be great. If you guys had a contributing guide and a link to a formal "issues" logger outside of GitHub (if that's the plan, e.g JIRA) that'd be great.

If these logs are not critical it'd be very much appreciated if a patch release was made in the interim since they are flooding our logs, in some scenarios our logs are only 20/100 lines, the other 80 are these db logs. We could downgrade to an older version of the module, but it'd be nice to avoid doing that for all of our apps.

matzew:
@evanshortiss adding something like this will help ? 
https://github.com/aerogear/aerogear-unifiedpush-server#project-info

Contributing guide... I can see if we can get something on feedhenry.org, that's similar to what other JBoss projects have

matzew:
@david-martin @evanshortiss I agree, that a proper logging would be best, and based on log level, the logging is verbose - or not 

@evanshortiss mind creating an FH JIRA ticket:
http://JIRa.jboss.org/browse/FH

evanshortiss:
@matzew yep, *project-info* be a good start. Personally, I think it'd be great to use the GitHub Issues page as much as possible, and just JIRA for internal sprints, etc.

Will open a ticket in JIRA, but would be great if this patch could be released in the meantime while the bigger log story is resolved 😄

matzew:
> GitHub Issues page as much as possible

most JBoss projects are using JIRA, and some have even disabled Issues :-)
That said, public community JIRAs are just fine.

MikeyBurkman:
@matzew Is this technically a JBOSS project? I don't see any reference to JBOSS anywhere on the RHMAP/FH site, nor do I see any reference to the platform (or really any non-JVM project) on the JBOSS page. 

(Plus there's no forum set up for RHMAP from what I can tell, which is where discussions like this are supposed to happen, according to the guidelines at https://developer.jboss.org/wiki/HelpBugReport)

evanshortiss:
Issue opened @matzew: FH-2708. I understand that's how JBoss has worked, but it'd make sense to evaluate that based on the community we're working within. Thanks for getting this stuff on the radar.

david-martin:
@evanshortiss I didn't mean for this PR to turn into a debate about process. Apologies for that.
@matzew I think this PR should be merged asap and released to unblock @evanshortiss. If there's a better place to have discussions about process like above, then lets move it there. Otherwise, it can continue here.

matzew:
replacing with 🍒 -pick on here: https://github.com/feedhenry/fh-mbaas-api/pull/7

matzew:
@evanshortiss for next sprint, @odra will be looking at all these JIRAs:
FH-2723, FH-2725 and FH-2726

PR #9, opened 2016-07-29
Removing feed API

matzew:
Done for [FH-2719](https://issues.jboss.org/browse/FH-2719)

@mmurphy @johnfriz  I found the above ticket on JIRA, and it seemed very reasonable to do

Wondering, since removing API - I feel version bump should be `5.14.0` ?


fh-automatron:

johnfriz:
@matzew - Since this is a breaking change to our API, it should probably be a major version bump, but don't really want to do this for a single API removal. I think we have a few other tickets knocking around from removing some other APIs. If we could bundle these together, then we could warrant a major version bump.

matzew:
@johnfriz  - +1 on `6.0.0` bump. What tickets do you have in mind ? we can likely include them in the next sprint(s)

johnfriz:
There are a number of APIs defined in https://github.com/feedhenry/fh-mbaas-api/blob/master/lib/api.js#L26-L50 that do not have any implementation (e.g. parse, stringify, utl) and can be removed. The "act" API has been replaced by the "service" api (they both use the same underlying implementation in act.js), so we could at least remove the definition of the "act" API. Would be good to review the others and ensure they are all documented. If not, we should either document them or remove them. We will need to announce this very prominently in out release notes for what ever release this goes out in, so people are aware of these breaking API changes.

If there are any other breaking change we are considering, I'd like to bundle them into the 6.0.0 release as well.

matzew:
Ok, @johnfriz sounds good - I also hear 6.0.0 feature branch :-)

matzew:
Ok, for now, we have `mbaas_6.0.0_feature` - and I will create some JIRAs out of your comments, @johnfriz

PR #22, opened 2016-10-03
$fh.host - Allow lookup of URL for GUIDs, not just current app URL

evanshortiss:
This PR adds a new backwards compatible change to $fh.hosts to enable users to get the URL of an app besides the current one.

The reason for the added dependency `fh-instance-url` is to facilitate parsing FH_SERVICE_MAP meaning $fh.host will work during local development scenarios.


lib/host.js
```diff
@@ -1,20 +1,21 @@
...
-var request = require('request');
+'use strict';
 
-module.exports = function host(cb) {
-  var url = 'https://' + process.env.FH_MILLICORE + '/box/srv/1.1/ide/apps/app/hosts';
-  var data = {
-    "guid": process.env.FH_INSTANCE,
-    "env": process.env.FH_ENV
-  }
+var process = require('process');
```
evanshortiss: Allows us to stub process during tests

lib/host.js
```diff
@@ -1,20 +1,21 @@
...
-    if (err) return cb(err);
+/**
+ * Returns the host URL for this app, or an application with the given guid
+ * @param  {String}   guid
+ * @param  {Function} cb
+ * @return {undefined}
+ */
+module.exports = function host (guid, cb) {
+  // Allow lookup of any guid on this domain
+  if (typeof guid === 'function') {
```
evanshortiss: Retains backwards compatibility.

fh-automatron:

evanshortiss:
@wei-lee @matzew thoughts? This is something people often need, to get the URL of the MBaaS to make a manual http(s) request

lib/host.js
```diff
@@ -1,20 +1,21 @@
...
+    cb = guid;
+    guid = process.env.FH_INSTANCE;
+  }
 
-    return cb(err, _.getPath(body, "hosts.url"));
-  });
-}
\ No newline at end of file
+  require('fh-instance-url')({
```
evanshortiss: Supports FH_SERVICE_MAP, but also does the GUID lookup if needed

wei-lee:
I have no problem with the code change here, but not sure if we should expose this function. Also it only works if both apps are in the same domain

wei-lee:
@evanshortiss what exactly the use case here?

evanshortiss:
@wei-lee sometimes we need to make a request to or get the URL of an App/Service on the domain. Right now the official SDK lacks that function so this will add it. I know it has been a need for a few apps we wrote so that's why _fh-instance-url_ exists. I know @tinyexplosions and @crosbie both need it too so having it in the core SDK would be nice, but naturally you guys hold the cards.

PR #27, opened 2016-11-24
Update tests section of readme

rachael-oregan:
## Motivation

Tests will return permission errors in test_fhsession if the redis password is not set correctly.


david-martin:
@rachael-oregan I've rebased this and will merge once build complete

PR #28, opened 2016-11-25
RHMAP-11703 - add new auth endpoint

rachael-oregan:
## Motivation

The studio preview for Web Apps (Embed Apps) incorrectly sends $fh.auth requests through the Web App and not through the domain

## Solution 

Add a new auth endpoint to fh-mbaas-api which determines if the app is running in local mode. In local mode the function provided by the user should be invoked.  In production mode the user's function should be proxied back to millicore.

### DIFF
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -58,18 +58,25 @@ var auth = {
    * correctly in local development/Studio Preview -
    * the js sdk dertermines envrironment mode from url string.
    * New endpoint determines environment mode from env var,
-   * sends request (req) and calls user provided function (cb).
+   * optionally sends request (localAuth - for local dev)
+   * and calls user provided function (cb).
    * @param  {object} req
-   * @param  {object} res
+   * @param  {function} localAuth - optional for local development.
    * @param  {object} cb
    */
-  performAuth: function (req, res, cb) {
+  performAuth: function (req, localAuth, cb) {
     // "FH_USE_LOCAL_DB" env var determines if the app is running in local mode.
     if (process.env.FH_USE_LOCAL_DB) {
-      // In local mode, invoke user function
-      return cb(undefined, res);
+      // In local mode, invoke user function/object
+      if (_.isFunction(localAuth)) {
+        return localAuth(req, cb);
+      } else if (_.isObject(localAuth)) {
+        return cb(undefined, localAuth);
+      } else {
+        return cb(undefined, {status: 200, body: {}});
+      }
     } else {
-      // In production mode, proxy request back to Core platform.
+            // In production mode, proxy request back to Core platform.
       return call("admin/authpolicy/auth", req.body, cb);
     }
   }
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -54,16 +54,16 @@ var auth = {
     }
   },
   /**
+   * New endpoint required as autentication ($fh.auth) requests not handled
+   * correctly in local development/Studio Preview -
+   * the js sdk dertermines envrironment mode from url string.
+   * New endpoint determines environment mode from env var,
+   * sends request (req) and calls user provided function (cb).
    * @param  {object} req
    * @param  {object} res
    * @param  {object} cb
-   * New endpoint required as autentication ($fh.auth) requests not handled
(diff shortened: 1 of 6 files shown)


lib/auth.js
```diff
@@ -52,6 +52,21 @@ var auth = {
     } else {
       return cb(null, false);
     }
+  },
+  /**
```
pb82: A short description of what it does and how it is used would be helpful.

pb82:
LGTM, maybe add a comment to the new function to explain why we need it and how it is used. I don't see any places where the new function is called though?

lib/auth.js
```diff
@@ -52,6 +52,25 @@ var auth = {
...
+   * @param  {object} req
+   * @param  {object} res
+   * @param  {object} cb
+   * New endpoint required as autentication ($fh.auth) requests were sending
```
fh-qube: MINOR Remove the useless trailing whitespaces at the end of this line. [rule](https://metrics.skunkhenry.com/coding_rules#rule_key=javascript%3ATrailingWhitespace)

fh-qube:
SonarQube analysis reported 2 issues:
* MINOR 2 minor

Watch the comments in this conversation to review them.

lib/auth.js
```diff
@@ -52,6 +52,26 @@ var auth = {
...
+  },
+  /**
+   * @param  {object} req
+   * @param  {object} res
+   * @param  {object} cb
+   * New endpoint required as autentication ($fh.auth) requests not handled
+   * correctly in local development/Studio Preview -
+   * the js sdk dertermines envrironment mode from url string.
+   * New endpoint determines environment mode from env var,
+   * sends request (req) and calls user provided function (cb).
```
wei-lee: normally in jsdoc, we put the description at the top, then parameter descriptions

lib/auth.js
```diff
@@ -52,6 +52,26 @@ var auth = {
...
+   * @param  {object} req
+   * @param  {object} res
+   * @param  {object} cb
+   */
+  getEnvironmentMode: function (req, res, cb) {
```
wei-lee: not sure about the name of the function, seems a bit confusing. Perhaps just call it `performAuth`?

lib/auth.js
```diff
@@ -52,6 +52,26 @@ var auth = {
...
+   * New endpoint required as autentication ($fh.auth) requests not handled
+   * correctly in local development/Studio Preview -
+   * the js sdk dertermines envrironment mode from url string.
+   * New endpoint determines environment mode from env var,
+   * sends request (req) and calls user provided function (cb).
+   * @param  {object} req
+   * @param  {object} res
+   * @param  {object} cb
+   */
+  performAuth: function (req, res, cb) {
```
wei-lee: it might be better to make the `res` parameter optional. Make it clear that it is only used in local mode in the comments. 

Also I am not sure if it's enough to only support `res` as a JSONObject. Sometimes users might need more complicated authentication process. I suggest to make it a function instead (and perhaps call it "localAuth") to make it clear.

Then the implementation could be like this:

```
if(typeof localAuth === 'function') {
  return localAuth(req, cb);
} else if(typeof localAuth === 'object') {
  return cb(null, localAuth);
} else {
  return cb(null, {});
}
```

lib/auth.js
```diff
@@ -52,6 +52,35 @@ var auth = {
...
+  performAuth: function (req, localAuth, cb) {
+    // "FH_USE_LOCAL_DB" env var determines if the app is running in local mode.
+    if (process.env.FH_USE_LOCAL_DB) {
+      // In local mode, invoke user function/object
+      if (_.isFunction(localAuth)) {
+        return localAuth(req, cb);
+      }
+      else if (_.isObject(localAuth)) {
+        return cb(undefined, localAuth);
+      }
```
fh-qube: MINOR Remove the useless trailing whitespaces at the end of this line. [rule](https://metrics.skunkhenry.com/coding_rules#rule_key=javascript%3ATrailingWhitespace)

fh-qube:
SonarQube analysis reported 1 issue:
* MINOR 1 minor

Watch the comments in this conversation to review them.

PR #30, opened 2017-01-03
ups version bump to fix redos vulnerability

witmicko:
# Issue
tough-cookie ReDoS via long string of semicolons
# Solution
use unifiedpush-node-sender version that uses patched version of request 

# JIRA
https://issues.jboss.org/browse/RAINCATCH-430


TommyJ1994:
Will be fixed in https://github.com/feedhenry/fh-mbaas-api/pull/99

PR #36, opened 2017-01-16
Convert extra sync tests to 'exports' and add to 'unit' test run

david-martin:
@evanshortiss Would you mind taking a look?
I converted the tests you added so they could be run alongside all the other existing tests.

1 minor change in the mocked out `sync` in `test_sync_bdd|handlers.js` was to stub out the callback for `stopAll` as its called in a global `after`.


test/test_sync_handlers.js
```diff
@@ -34,7 +32,7 @@ describe('$fh.sync', function () {
       init: sinon.spy(),
       createDataset: sinon.spy(),
       stopDatasetSync: sinon.spy(),
-      stopAllDatasetSync: sinon.spy(),
+      stopAllDatasetSync: sinon.stub().callsArgAsync(0),
```
david-martin: @evanshortiss this change ensures the global `after` will have its callback called when it does `fh.sync.stopAll`

evanshortiss:
Weird, I thought they ran fine with the reset of the tests when I did the original PR. LGTM though 👍

PR #38, opened 2017-01-17
[WIP] FH-3023 Current progress

aidenkeating:
* Using `async-json` library.
  * Update dependent functions to be async

@david-martin This is the current progress on the task.

### DIFF
--- a/lib/sync-DataSetModel.js
+++ b/lib/sync-DataSetModel.js
@@ -259,19 +259,24 @@ var self = {
 
     self.getDataset(dataset_id, function (err, dataset) {
       if (err) return err;
-      var clientHash = self.getClientHash(query_params, meta_data);
-      var datasetClient = dataset.clients[clientHash];
-      if (!datasetClient) {
-        self.createDatasetClient(dataset_id, query_params, meta_data, function (err, dsc) {
-          if (err) return cb(err);
-          recordAccessInfo(dsc, params);
-          return cb(null, dsc);
-        });
-      }
-      else {
-        recordAccessInfo(datasetClient, params);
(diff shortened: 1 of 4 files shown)


wtrocki:
ping @wtrocki

lib/sync-DataSetModel.js
```diff
@@ -349,11 +354,16 @@ var self = {
     syncUtil.doLog(datasetClient.datasetId, 'silly', 'Now there are ' + datasetClient.pendingCallbacks.length + ' calbacks for datasetclient id =' + datasetClient.id, datasetClient);
   },
 
-  getClientHash: function (query_params, meta_data) {
-    var queryParamsHash = syncUtil.generateHash(query_params);
-    var metaDataHash = syncUtil.generateHash(meta_data);
+  getClientHash: function (query_params, meta_data, cb) {
```
wtrocki: If async module is included in this module we can use parallel call.

lib/sync-DataSetModel.js
```diff
@@ -349,11 +354,16 @@ var self = {
...
+    syncUtil.generateHash(query_params, function(err, queryParamsHash) {
+      if (err) return cb && cb(err);
+
+      syncUtil.generateHash(meta_data, function(err, metaDataHash) {
```
wtrocki: [Suggestion] Another improvement here would be to figure out where this hashes are used and instead of hashing twice maybe we can just append both together and generate single hash. Not sure how much changes this would require on client, but leaving this as suggestion.

wtrocki:
@aidenkeating What happened here?

PR #42, opened 2017-01-27
FH-3067 Adding test_sync-serv.js

danbev:
The goal of this task was to investigate how we can avoid doing a list before deleting acknowledgements. @david-martin already did the heavy lifting here adding this functionality and this PR is mainly adding a test and getting the complete test suite to pass.

JIRA: https://issues.jboss.org/browse/FH-3067

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -4,27 +4,27 @@
   "dependencies": {
     "async": {
       "version": "0.2.9",
-      "from": "async@0.2.9",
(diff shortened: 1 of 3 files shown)


david-martin:
@danbev I'm good with merging this as its not to feedhenry:master.
We should look at re-using the MongoClient in fh-db rather than having a separate mongdb connection before merging to feedhenry:master (as per https://issues.jboss.org/browse/FH-3084)

PR #47, opened 2017-02-13
RHMAP-12825 - Env variables instead of hard-coded https protocol and port

trepel:
**Motivation**

fh-mbaas-api cannot be fully used for clusters that doesn't support https protocol (e.g. ENG1).

**Modifications**

The biggest change is in hosts.js because the cfg object needed to be passed along. Otherwise I just replaced the hard coded https related stuff so that http can be used as well.

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -45,7 +45,7 @@ function FHapi(cfg, logr) {
     service: require('./act')(cfg),
     sec: sec.security,
     auth: require('./auth')(cfg),
-    host: require('./host'),
+    host: require('./host')(cfg),
     permission_map: require('fh-db').permission_map,
     hash: function (opts, callback) {
       var p = {
@@ -87,11 +87,13 @@ function FHapi(cfg, logr) {
 }
 
 /**
- * Initilisation returns the $fh object to clients
+ * Initialisation returns the $fh object to clients
  */
 module.exports = (function () {
-  // First setup the required config params from  env variables
+  // First setup the required config params from env variables
   var millicore = process.env.FH_MILLICORE || 'NO-MILLICORE-DEFINED';
+  var millicore_protocol = process.env.FH_MILLICORE_PROTOCOL || 'https';
+  var millicore_port = process.env.FH_MILLICORE_PORT || 443;
   var domain = process.env.FH_DOMAIN || 'NO-DOMAIN-DEFINED';
   var instance = process.env.FH_INSTANCE || 'NO-INSTANCE-DEFINED';
   var appname = process.env.FH_APPNAME || 'NO-APPNAME-DEFINED';
@@ -143,8 +145,9 @@ module.exports = (function () {
   var cfg = {
     fhapi: {
       appname: appname,
+      protocol: millicore_protocol,
       millicore: millicore,
-      port: 443,
+      port: millicore_port,
       domain: domain,
       instance: instance,
       widget: widget,
--- a/lib/call.js
+++ b/lib/call.js
@@ -1,6 +1,7 @@
 var config,
   _ = require('underscore'),
   https = require('https'),
+  http = require('http'),
   futils = require('./fhutils'),
   fhutils;
 module.exports = function (cfg) {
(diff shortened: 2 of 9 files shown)


lib/host.js
```diff
@@ -1,20 +1,30 @@
...
-    json: true,
-    body: data
-  }, function (err, res, body) {
-    if (err) return cb(err);
+  var fhutils = new futils(cfg);
+  var millicoreProps = fhutils.getMillicoreProps();
 
-    return cb(err, _.getPath(body, "hosts.url"));
-  });
+  return function host(cb) {
```
fh-qube: BLOCKER Add a semicolon at the end of this statement. [rule](https://metrics.skunkhenry.com/coding_rules#rule_key=javascript%3ASemicolon)

fh-qube:
SonarQube analysis reported 1 issue:
* BLOCKER 1 blocker

Watch the comments in this conversation to review them.

lib/host.js
```diff
@@ -1,20 +1,30 @@
...
+var futils = require('./fhutils');
 
-module.exports = function host(cb) {
-  var url = 'https://' + process.env.FH_MILLICORE + '/box/srv/1.1/ide/apps/app/hosts';
-  var data = {
-    "guid": process.env.FH_INSTANCE,
-    "env": process.env.FH_ENV
+module.exports = function(cfg) {
+  if (!cfg) {
+    return;
```
sedroche: If `cfg` is required for the module to be instantiated maybe throw an error rather than silently returning?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fh-mbaas-api",
-  "version": "6.1.5",
+  "version": "6.1.6",
```
sedroche: you should update sonarqube `.properties` file and shrinkwrap as well

lib/host.js
trepel: @sedroche looking at the other modules, they either don't care, or have this ugly piece of code, or use `assert.ok(cfg, 'cfg is undefined');` Does this assert approach looks better to you ?

package.json
trepel: @sedroche thanks, I am not familiar with what needs to be done along with verison bump. Will update

lib/host.js
sedroche: Yes the assert approach looks good to me

lib/call.js
```diff
@@ -35,12 +35,13 @@ var call = function call(path, params, callback) {
   addParams["widget"] = props.widgId;
 
   var fhResp = {};
-  var req = https.request(options, function (res) {
+  var protocolLibrary = props.protocol === 'http' ? http : https;
```
pb82: What about something like `protocolLibrary = require(props.protocol);`? Then you don't need to require both of them.

test/test_fhpush.js
```diff
@@ -26,6 +26,7 @@ describe('$fh.push', function () {
         widget: 'a',
         instance: 'b',
         millicore: 'fake-domain.feedhenry.com',
+        protocol: 'https',
```
pb82: Maybe add a test that uses `http` just to make sure the code path work the same with http.

lib/api.js
```diff
@@ -87,11 +87,13 @@ function FHapi(cfg, logr) {
...
 
 /**
- * Initilisation returns the $fh object to clients
+ * Initialisation returns the $fh object to clients
  */
 module.exports = (function () {
-  // First setup the required config params from  env variables
+  // First setup the required config params from env variables
   var millicore = process.env.FH_MILLICORE || 'NO-MILLICORE-DEFINED';
+  var millicore_protocol = process.env.FH_MILLICORE_PROTOCOL || 'https';
```
pb82: If these are new env vars we should document them somewhere, probably in the README.

david-martin:
@trepel I think we can merge these changes if you rebase and add a test (as suggested by @pb82)

david-martin:
@trepel I'm closing this due to being out of date.
If you want to proceed with these changes, please open a new PR.

PR #48, opened 2017-02-13
remove references to the datasetClientRecords

wei-lee:
## Motive

The `datasetClientRecords` is not used anywhere. We should remove it to improve the memory usage of the sync framework.

## Changes

Remove the references to the `datasetClientRecords` variable and methods.

## JIRA

https://issues.jboss.org/browse/FH-3124

ping @david-martin @aidenkeating

### DIFF
--- a/lib/sync-DataSetModel.js
+++ b/lib/sync-DataSetModel.js
@@ -40,7 +40,6 @@ var self = {
   },
 
   datasets: {},
-  datasetClientRecords: {},
   deletedDatasets: {},
   createDatasetOnDemand: true,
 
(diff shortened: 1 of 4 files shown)


lib/sync-DataSetModel.js
```diff
@@ -527,40 +510,19 @@ var self = {
   toJSON: function (dataset_id, returnData, cb) {
```
wei-lee: @david-martin I think it's better to keep the `toJSON` method as it can be useful for debugging purpose.

david-martin:
Closing this. See https://issues.jboss.org/browse/FH-3124

PR #53, opened 2017-02-16
support custom hashing rules for sync records

evanshortiss:
The idea behind this PR is that our default sync hashing is extremely robust and reliable, but the tradeoff is that it's expensive. This is due to the fact it recursively traverses an entire object, sorts the properties in each sub object/array, stringifies them, and then computes a hash using the sorted objects.

This expensive algorithm is not always necessary. For example, we might have a dataset that contains records like so:

```js
// Note Schema
{
  uid: String
  updateTs: Date,
  updateCount: Number
  title: String,
  content: String,
  meta: {
    // Could be many of these
    [key]: Object
  }
}
```

In this case instead of hashing the entire object we could simply hash the `uid+ updateTs.toString() + updateCount` or similar since each time a Note is updated/modified we would change these.

I tested locally the time it would take to complete the hashing for 500 `list` calls to the sync and found the following:

* Current Sync Hashing: 1259ms
* Custom Sync Hashing: 115ms

The test used node v4.4.3 and a ~90K piece of production JSON. I've seen the production JSON can go up to 250K but thought ~90K was a nice middle ground.

We have to support around 2000 users syncing 4 collections so this could make a real difference to throughput.

### DIFF
--- a/lib/sync-DataSetModel.js
+++ b/lib/sync-DataSetModel.js
@@ -89,7 +89,7 @@ var self = {
       for (var i in records) {
         var rec = {};
         var recData = records[i];
-        var hash = hashFn(recData);
+        var hash = hashFn(recData, i); // pass the record and its id
         hashes.push(hash);
         rec.data = recData;
         rec.hash = hash;
--- a/lib/sync-DataSetModel.js
+++ b/lib/sync-DataSetModel.js
@@ -463,37 +463,7 @@ var self = {
           }
         }
       }
-    )
-
-    self.doListHandler(dataset.id, datasetClient.queryParams, function (err, records) {
-      if (err) {
-        datasetClient.syncRunning = false;
-        datasetClient.syncLoopEnd = new Date().getTime();
-        return cb && cb(err);
-      }
-
-      self.generateRecordHashes(dataset.id, records, function (err, hashes) {
-        if (err) {
-          datasetClient.syncRunning = false;
-          datasetClient.syncLoopEnd = new Date().getTime();
-          return cb && cb(err);
-        }
-
-        var globalHash = hashes.hash;
-
-        var previousHash = datasetClient.dataHash ? datasetClient.dataHash : '<undefined>';
-        syncUtil.doLog(dataset.id, 'verbose', 'doSyncList cb ' + ( cb !== undefined) + ' - Global Hash (prev :: cur) = ' + previousHash + ' ::  ' + globalHash);
-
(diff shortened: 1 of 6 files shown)


wei-lee:
@evanshortiss This is really good work Evan and thanks again for the contribution. I don't think there is any problem with your changes here, but I do think there are a few more places you need to apply the same hash function. For example here:

https://github.com/feedhenry/fh-mbaas-api/blob/24e795bd2671674cd9cc770b020c268cd6cc4b75/lib/sync-srv.js#L302

So basically take a look and see where `syncUtil.generateHash` is being used and some of them should be replaced with the right hash function. This will help to make sure the change will be backward compatible.

Does this make sense?

evanshortiss:
@wei-lee will get this updated soon now that I'm back from PTO 👍

evanshortiss:
@wei-lee updated. It now also passes the `uid` since that would be very useful in generating collision resistant hashes.

evanshortiss:
This has been tested with a sample app. I added the below lines [here](https://github.com/RHMAP-Sample-Mobile-Apps/sync-angular-cloud/blob/master/application.js#L40) and ran the accompanying client and service.

```js
  // Use a custom hashing function for records
  mbaasApi.sync.handleHash('users', function (record, uid) {
    var hasher = crypto.createHash('md5');

    // Simply hash based on these strings
    hasher.update(record.firstname + record.lastname + uid);

    var ret = hasher.digest('hex');

    log.trace(`generated hash "${ret}" for sync entry "${uid}". data:`, record);

    return ret;
  });
```

Using the above setup I tested LIST, CREATE, UPDATE, and DELETE. This also indirectly tests READ.

lib/sync-DataSetModel.js
```diff
@@ -375,43 +429,70 @@ var self = {
...
+          return cb && cb(err);
+        } else {
+          if (cb) {
+            cb(null, hashes);
+          }
+        }
+      }
+    )
+
     self.doListHandler(dataset.id, datasetClient.queryParams, function (err, records) {
```
wei-lee: @evanshortiss I think we should remove this function? Looks like it's a duplication of what you have added above?

wei-lee:
@evanshortiss thanks for the update. It looks good now. Just one question. It looks like we have some duplicated code in one of the functions

lib/sync-DataSetModel.js
evanshortiss: Ah, yep. I think I left it in while testing!

PR #56, opened 2017-02-21
Fh 3142 sync scheduler

wei-lee:
## Motive

As described in https://issues.jboss.org/browse/FH-3142, we need a scheduler that will check the dataset clients periodically and push sync requests on the queue is a dataset client is due to sync. This PR will add the implementation for the scheduler.

Some changes are also done on the sync request processor to make sure the right information about the sync loop is being recorded.

ping @aidenkeating @danbev for code review. Not sure if you guys have see this doc: https://docs.google.com/document/d/1s6DkFE2P5dMLto7FduNxZgspUO221kuBAQXUCW45tDU/edit#heading=h.ddrmquooxfvy. This PR implements the "sync_scheduler" part in the diagram.


lib/sync/DatasetClient.js
```diff
@@ -0,0 +1,107 @@
...
+    return true;
+  }
+  // Time between sync loops has passed - do another sync
+  var lastSyncCmp = self.props.syncLoopEnd;
+  if (!lastSyncCmp) {
+    //the sync run hasn't been finished before, can't decide
+    return false;
+  }
+  var syncFrequency = self.config.syncFrequency * 1000;
+  var timeSinceLastSync = new Date().getTime() - lastSyncCmp;
```
aidenkeating: This could be `Date.now()` I think. I don't know if it's faster or anything but it might make it look a bit simpler. There's a few other places this could be replaced also.

```javascript
var timeSinceLastSync = Date.now() - lastSyncCmp;
```

lib/sync/sync-scheduler.js
```diff
@@ -1,3 +1,140 @@
...
+      metricsClient.gauge(metrics.KEYS.SYNC_SCHEDULER_CHECK_TIME, {success: false}, timing);
+      return cb(err);
+    }
+
+    var datasetClientsToSync = [];
+    var datasetClientsToRemove = [];
+
+    _.each(datasetClients, function(datasetClientJson){
+      var datasetClient = DatasetClient.fromJSON(datasetClientJson);
+      var dataset_id = datasetClient.getDatasetId();
```
aidenkeating: I think this can be `datasetId`

lib/sync/sync-scheduler.js
```diff
@@ -1,3 +1,140 @@
...
+
+/**
+ * Loop through all the dataset clients and see if any of them is due to sync with the backend, or if the dataset is not active anymore and should be removed.
+ * @param {Function} cb the callback function
+ */
+SyncScheduler.prototype.checkDatasetsForSyncing = function(cb) {
+  var self = this;
+  var timer = metrics.startTimer();
+  syncStorage.listDatasetClients(function(err, datasetClients){
+    var timing;
```
aidenkeating: Do we need this? I think we can just use `timer.stop()` in the two places it's needed

lib/sync/sync-processor.js
```diff
@@ -46,19 +25,86 @@ function syncWithBackend(payload, callback) {
...
+
+  async.waterfall([
+    function setSyncStart(cb) {
+      var syncLoopStartTime = new Date().getTime();
+      syncStorage.updateDatasetClient(datasetClientId, {syncLoopStart: syncLoopStartTime, syncLoopEnd: syncLoopStartTime + expectedTimeout}, function(err){
+        return cb(err);
       });
+    },
+    function listData(cb) {
+      listDataWithTimeout({datasetClientId: datasetClientId, datasetId: datasetId, queryParams: queryParams, metaData: metaData}, expectedTimeout - 2*1000, cb);
```
aidenkeating: I have probably missed something obvious, but why are we taking away the 2 seconds here?

lib/sync/sync-processor.js
```diff
@@ -46,19 +25,86 @@ function syncWithBackend(payload, callback) {
...
+  if (!datasetClientId || !datasetId) {
+    recordProcessTime(startTime, false);
+    syncUtil.doLog(syncUtil.SYNC_LOGGER, "error", "no datasetId value found in sync request payload" + util.inspect(payload));
+    return callback();
+  }
+
+  var queryParams = payload.queryParams || {};
+  var metaData = payload.metaData || {};
+  //we need to add this so that if this sync processor crashed before reaching the end, the scheduler will still be able to push a sync request
+  var expectedTimeout = datasets.getDatasetConfig(datasetId).backendListTimeout * 1000 || 5*60*1000;
```
aidenkeating: Another question, why is this defaulting to 5 minutes? Could it be configurable? If not, maybe it should be defined as a constant? Wdyt?

lib/sync/DatasetClient.js
wei-lee: Yes, you are right. I am just used to use `new Date().getTime()`. Will change it

lib/sync/sync-processor.js
wei-lee: Yes, it will be configurable. Developers will be able to override this value at per-dataset level.

lib/sync/sync-processor.js
wei-lee: yeah, we don't need that anymore. I used to have a lock here and the 2 seconds is used to make sure we have enough time to process the rest of the function before releasing it. Then I realise the lock may not be required, but forgot to change this back.

lib/sync/sync-scheduler.js
wei-lee: you are right, we don't. Will change this.

PR #61, opened 2017-02-23
FH-3171 Add hash provider for sync

aidenkeating:
Adds a module that allows for the hashing algorithm used for each
dataset to be overridden. For each dataset there is a `recordHash`
and a `globalHash` function. 

The `recordHash` function deals with
the hashing of single records in a dataset. This can be overridden
by calling `setRecordHash` and passing in the ID of the dataset.

The `globalHash` function deals with hashing for an entire dataset,
it accepts an array of record hashes and uses these to calculate
a hash for the entire dataset. This can be overridden in a similar
way to the record hash.

### DIFF
--- a/lib/sync/hashProvider.js
+++ b/lib/sync/hashProvider.js
@@ -1,18 +1,89 @@
-module.exports = function() {
-  return {
-    /**
-     * @param datasetId
-     * @param record
-     */
-    recordHash: function() {
-      //TODO: implement this
-    },
-    /**
-     * @param datasetId
-     * @param hashesArrayOfAllRecords
-     */
-    globalHash: function() {
-      //TODO: implement this
-    }
-  };
+var syncUtil = require('./util');
+
+/**
+ * Store for the overridden hashing functions for each dataset.
+ */
+var overrides = {
+  record: {},
+  global: {}
+}
+
+/**
+ * Set the default hashing algorithm.
+ */
+var defaultHashFn = syncUtil.generateHash;
+
+/**
+ * Invoke the datasets record hash function on a record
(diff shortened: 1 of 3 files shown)


aidenkeating:
@wei-lee @danbev Would you mind taking a look please?

wei-lee:
👀

lib/sync/hashProvider.js
```diff
@@ -1,18 +1,91 @@
...
+var defaultHashFn = syncUtil.generateHash;
+
+/**
+ * Invoke the datasets record hash function on a record
+ * @param {String} datasetId the id of the dataset
+ * @param {Object} record record in the dataset to hash
+ * @returns {String} the resulting record hash
+ */
+function recordHash(datasetId, record) {
+  var hashFn = overrides.record[datasetId] || defaultHashFn;
```
wei-lee: just DRY, you can use `getRecordHashFn` here

lib/sync/hashProvider.js
```diff
@@ -1,18 +1,91 @@
...
+}
+
+/**
+ * Invoke the datasets global hash function on an array of hashes
+ * @param {String} datasetId the id of the dataset
+ * @param {String[]} hashes array of record hashes in the dataset
+ * @returns {String} the resulting global hash
+ */
+function globalHash(datasetId, hashes) {
+  var hashFn = overrides.global[datasetId] || defaultHashFn;
```
wei-lee: same as above, you can use `getGlobalHashFn`

aidenkeating:
@wei-lee Thanks! Made those changes

PR #62, opened 2017-02-24
implement dataHandler

danbev:
Motivation:
Implement the datahandlers so that each dataset can have it's own
datahandler for list/create/read/update/delete/collision.

JIRA:
https://issues.jboss.org/browse/FH-3170


danbev:
Thanks @wei-lee

PR #63, opened 2017-02-24
Fh 3150 implement sync records

wei-lee:
Add the implementation for the syncRecords API.

The tricky part here is that we want to use the data that is saved locally in mongodb for performance reasons. However, this data is only get updated at the end of every sync loop, which means it may not match the backed data. The data returned by the endpoint will be treated as the current backend data and it will cause the client to remove/delete/update data. This means if there are some changes applied after the last sync loop completed, or waiting for be processed, they will not be in the local db, and the client will be asked to remove those data (then get it back again after the next sync loop run).

To avoid this problem, we will try to list those applied changes and any other pending changes for a given client, and remove those records from both client and the current local database records. This means those records will not be compared and the client will then keep its own changes until the next sync loop.

ping @aidenkeating @danbev  for review


lib/sync/api-syncRecords.js
```diff
@@ -0,0 +1,230 @@
...
+ * Sync the data records for the given client
+ * @param {String} datasetId 
+ * @param {Object} params the request body, it normally contain those fields:
+ * @param {Object} params.query_params the query parameter for the dataset from the client
+ * @param {Object} params.meta_data the meta data for the dataset from the client
+ * @param {Object} params._fh an object added by the client sdk. it should have the `cuid` of the client
+ * @param {Object} params.clientRecs the client records
+ * @param {Function} cb
+ */
+function syncRecoreds(datasetId, params, cb) {
```
aidenkeating: I think this is a typo, `syncRecoreds` should be `syncRecords`, same in the exports

wei-lee: @aidenkeating yes, proves again I can't type 😺 . Will fix this. Thanks.

PR #64, opened 2017-02-27
add implementation for default dataHandler

danbev:
Motivation:
We need to have a default implementation of the dataHandler for cases
where a user does not provide an override.

Modifications:
Added the default dataHandler implementation and updated the
dataHandlersModule to take the defaults as its options so that the
defaults will always be used unless overridden.

JIRA:
https://issues.jboss.org/browse/FH-3177

### DIFF
--- a/lib/sync/default-dataHandlers.js
+++ b/lib/sync/default-dataHandlers.js
@@ -20,13 +20,7 @@ module.exports = function (db) {
         if (err) {
           return cb(err);
         }
-        var data = {};
-        array.forEach(function extractUidAndData(value) {
-          var uid = value._id;
-          delete value._id;
-          data[uid] = value;
-        });
-        cb(null, data);
+        cb(null, toObject(array));
       });
     },
 
@@ -71,14 +65,24 @@ module.exports = function (db) {
                      'doCollision : ' + dataset_id +
                      ' :: collisionFields=' + util.inspect(collisionFields) +
                      ' :: meta=' + util.inspect(meta_data));
-      mongo.collection(collisionCollection(dataset_id)).insertOne(collisionFields, cb);
+      mongo.collection(collisionCollection(dataset_id)).insertOne(collisionFields, function(err, res) {
+        if (err) {
+          return cb(err);
+        }
+        cb(null, makeResponse(res.ops[0]));
+      });
     },
 
     listCollisions: function (dataset_id, meta_data, cb) {
       syncUtil.doLog(dataset_id, 'verbose',
                      'listCollisions : ' + dataset_id +
                      ' :: meta=' + util.inspect(meta_data));
-      mongo.collection(collisionCollection(dataset_id)).find().toArray(cb);
+      mongo.collection(collisionCollection(dataset_id)).find().toArray(function (err, array) {
+        if (err) {
+          return cb(err);
+        }
+        cb(null, toObject(array));
+      });
     },
 
     removeCollision: function (dataset_id, uid, meta_data, cb) {
@@ -92,6 +96,17 @@ module.exports = function (db) {
   };
 
 };
+
+function toObject(array) {
+  var data = {};
+  array.forEach(function extractUidAndData(value) {
+    var uid = value._id;
+    delete value._id;
+    data[uid] = value;
+  });
+  return data;
+}
+
 
 function makeResponse(res) {
   var data = {
--- a/lib/sync/default-dataHandlers.js
+++ b/lib/sync/default-dataHandlers.js
@@ -23,7 +23,12 @@ module.exports = function (db) {
       syncUtil.doLog(dataset_id, 'verbose',
                      'doCreate : ' + dataset_id + ' :: ' + util.inspect(data) +
                      ' :: meta=' + util.inspect(meta_data));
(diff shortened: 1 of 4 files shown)


danbev:
@wei-lee Would you mind taking a look and see what you think?

wei-lee:
👀

lib/sync/default-dataHandlers.js
```diff
@@ -0,0 +1,79 @@
...
+                     ' :: query_params=' + util.inspect(params) +
+                     ' :: meta_data=' + util.inspect(meta_data));
+      mongo.collection(dataset_id).find(params).toArray(cb);
+    },
+
+    doCreate: function (dataset_id, data, meta_data, cb) {
+      syncUtil.doLog(dataset_id, 'verbose',
+                     'doCreate : ' + dataset_id + ' :: ' + util.inspect(data) +
+                     ' :: meta=' + util.inspect(meta_data));
+      mongo.collection(dataset_id).insertOne(data, cb);
```
wei-lee: I believe the `create` implementation have to at least return the `uid` (the `_id`) field as part of the callback.  See https://github.com/feedhenry/fh-mbaas-api/blob/master/lib/sync-datahandler.js#L47

lib/sync/default-dataHandlers.js
```diff
@@ -0,0 +1,79 @@
...
+                      ' :: meta=' + util.inspect(meta_data));
+      mongo.collection(dataset_id).findOne({"_id" : uid}, cb);
+    },
+
+    doUpdate: function (dataset_id, uid, data, meta_data, cb) {
+      syncUtil.doLog(dataset_id, 'verbose',
+                     'doUpdate : ' + dataset_id +
+                     ' :: ' + uid + ' :: ' + util.inspect(data) +
+                     ' :: meta=' + util.inspect(meta_data));
+      mongo.collection(dataset_id).update({"_id" : uid}, data, cb);
```
wei-lee: looks like `update` is being deprecated. Might be better using `updateOne` instead.

lib/sync/default-dataHandlers.js
```diff
@@ -0,0 +1,79 @@
...
+                     ' :: meta=' + util.inspect(meta_data));
+      mongo.collection(dataset_id).remove({"_id" : uid}, cb);
+    },
+
+    handleCollision: function (dataset_id, meta_data, collisionFields, cb) {
+      syncUtil.doLog(dataset_id, 'verbose',
+                     'doCollision : ' + dataset_id +
+                     ' :: collisionFields=' + util.inspect(collisionFields) +
+                     ' :: meta=' + util.inspect(meta_data));
+      mongo.collection(dataset_id + COLL_POSTFIX).insertOne(collisionFields, cb);
```
wei-lee: I think it might better to move the collection name to a function. This way it will be easier to change it if we need to.

test/sync/test_default-dataHandlers.js
```diff
@@ -0,0 +1,178 @@
...
+var sinon = require('sinon');
+var defaultDataHandlersModule = require('../../lib/sync/default-dataHandlers.js');
+
+var id = 'datahandler_test';
+var queryParams = {};
+var metaData = {};
+
+// stubs when there is no real MongoDB
+var collectionStub = {
+  insertOne: sinon.stub().callsArgWith(1, null, {
```
wei-lee: `callsArgWith` is ok. You can also use sinon.yields, which will automatically pass the give parameters to callback function. A bit easier to use.

lib/sync/default-dataHandlers.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+  mongo = db
+
+  return {
+    doList: function (dataset_id, params, meta_data, cb) {
+      syncUtil.doLog(dataset_id, 'verbose',
+                     'doList : ' + dataset_id +
+                     ' :: query_params=' + util.inspect(params) +
+                     ' :: meta_data=' + util.inspect(meta_data));
+      mongo.collection(dataset_id).find(params).toArray(cb);
```
wei-lee: I think for the list operation, the sync framework expected the data return in a object, and the key should be the uid of each record. See https://access.redhat.com/documentation/en-us/red_hat_mobile_application_platform_hosted/3/html/cloud_api/fh-sync#fh-sync-handlelist

lib/sync/default-dataHandlers.js
wei-lee: See https://access.redhat.com/documentation/en-us/red_hat_mobile_application_platform_hosted/3/html/cloud_api/fh-sync#fh-sync-handlecreate

lib/sync/default-dataHandlers.js
danbev: Ah, I was not aware of that. I'll fix that.

lib/sync/default-dataHandlers.js
danbev: Let me take a look at that and add this.

lib/sync/default-dataHandlers.js
danbev: Yeah good point, let me fix that.

danbev:
@wei-lee Thanks for the review! I think I've covered all your change requests now. Please take a look when you get time.

PR #65, opened 2017-02-27
FH-3174 Implement the lock interface

aidenkeating:
This implements a wrapper around a modified version of mongodb-lock
which supports 2.x of the mongodb client.

The wrapper stores all locks created and will create a lock if it
doesn't exist during an acquisition request.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -43,7 +43,8 @@ module.exports = function(grunt) {
     './test/sync/test_dataHandlers.js',
     './test/sync/test_api-syncRecords.js',
     './test/sync/test_default-dataHandlers.js',
-    './test/sync/test_interceptors.js'
+    './test/sync/test_interceptors.js',
+    './test/sync/test_lock.js'
   ];
   var unit_args = _.map(tests, makeTestArgs);
   var test_runner = '_mocha';
--- a/lib/sync/lock.js
+++ b/lib/sync/lock.js
@@ -1,23 +1,31 @@
-// TODO Use mongodb for locking via https://github.com/chilts/mongodb-lock
-// TODO when using https://github.com/chilts/mongodb-lock, ensure old locks are removed
+var mongoLock = require('mongodb-lock');
 
+var locks = {};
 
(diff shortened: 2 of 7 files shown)


lib/sync/lock.js
```diff
@@ -1,23 +1,31 @@
-// TODO Use mongodb for locking via https://github.com/chilts/mongodb-lock
-// TODO when using https://github.com/chilts/mongodb-lock, ensure old locks are removed
+var mongoLock = require('@akeating-redhat/mongodb-lock');
```
wei-lee: I don't think this will work for node 0.10.

I think you can create a PR to the original module to upgrade the mongodb. Based on @david-martin experience, the author should happy to accept the PR.

lib/sync/lock.js
```diff
@@ -1,23 +1,31 @@
...
 
+module.exports = function(mongoClient, collectionName) {
+  collectionName = collectionName || 'sync_locks';
+  return {
+    locks: {},
```
wei-lee: very small thing but I don't think the `locks` variable and the `findOrCreate` function need to be public? They could be private functions.

aidenkeating:
@wei-lee Thanks for the review.

Here's the PR for the changes to `mongodb-lock`, I've made them backwards compatible now: https://github.com/chilts/mongodb-lock/pull/4

I'll make those other changes now

aidenkeating:
retest this please

aidenkeating:
retest this please

aidenkeating:
retest this please

Gruntfile.js
```diff
@@ -43,7 +43,7 @@ module.exports = function(grunt) {
     './test/sync/test_dataHandlers.js',
     './test/sync/test_api-syncRecords.js',
     './test/sync/test_default-dataHandlers.js',
-    './test/sync/test_interceptors.js'
+    './test/sync/test_lock.js'
```
wei-lee: this should be a new test, rather than changing an existing one?

PR #71, opened 2017-03-01
Fix various issues discovered when attempting to run acceptance tests.

david-martin:
Call `sync.start()` to ensure the sync scheduler is active whenever a
client connects.
Fix datahandler wiring of function names.
Fix db client passing into the MongoDB Queue module.

### DIFF
--- a/lib/sync/dataHandlers.js
+++ b/lib/sync/dataHandlers.js
@@ -161,14 +161,14 @@ module.exports = function(options) {
   };
 
   function Handler(defaultHandlers) {
-    this.listHandler = defaultHandlers.listHandler;
-    this.createHandler = defaultHandlers.createHandler;
-    this.readHandler = defaultHandlers.readHandler;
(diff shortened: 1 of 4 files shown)


lib/sync/dataHandlers.js
```diff
@@ -186,6 +186,7 @@ module.exports = function(options) {
       handler = new Handler(options.defaultHandlers);
       handlers[datasetId] = handler;
     }
+    console.log('options.defaultHandlers', options.defaultHandlers);
```
wei-lee: can we remove this console log

lib/sync/index.js
```diff
@@ -67,7 +67,10 @@ function invoke(dataset_id, params, callback) {
   }
   var fnHandler = module.exports[fn];
 
-  return fnHandler(dataset_id, params, callback);
+  start(function(err) {
+    if (err) return callback(err);
```
wei-lee: just to be consistent, can we move the return statement to a newline like everywhere else

PR #72, opened 2017-03-01
FH-3198 Test collision handlers

aidenkeating:
This adds tests for overriding the collision handlers and the default
collision handlers.

This also includes fixes for the collision handling itself that were
uncovered during testing.

### DIFF
--- a/integration/sync/test_collision-handlers.js
+++ b/integration/sync/test_collision-handlers.js
@@ -1,0 +1,118 @@
+var MongoClient = require('mongodb').MongoClient;
+var assert = require('assert');
+var _ = require('underscore');
+var defaultDataHandlersModule = require('../../lib/sync/default-dataHandlers');
+var dataHandlersModule = require('../../lib/sync/dataHandlers');
+
+var DATASETID = "collisionHandlersTest";
+var MONGODB_URL = "mongodb://127.0.0.1:27017/test";
+
+var mongodb;
+var dataHandlers;
+
+function getCollisionData() {
+  return {
+    pre: { test: 'test1' },
+    post: { test: 'test2' },
+    hash: 'testhash',
+    ts: Date.now(),
+    uid: 'testUid'
+  };
+}
+
+module.exports = {
+  'before': function(done) {
+    MongoClient.connect(MONGODB_URL, function(err, db){
+      if (err) {
+        console.log('mongodb connection error', err);
+        return done(err);
+      }
+      mongodb = db;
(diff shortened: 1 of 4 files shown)


aidenkeating:
I think most of the custom overrides here can be stubbed to make the tests here simpler

aidenkeating:
@david-martin @wei-lee @danbev Would you mind taking a look?

I'm also noticing a problem with the storage tests locally, well probably with the storage implementation. When set is an empty object the tests are failing, it might be the version of mongo I'm using though.

integration/sync/test_collision-handlers.js
```diff
@@ -0,0 +1,115 @@
...
+    pre: { test: 'test1' },
+    post: { test: 'test2' },
+    hash: 'testhash',
+    ts: Date.now(),
+    uid: 'testUid'
+  };
+}
+
+module.exports = {
+  'before': function(done) {
```
wei-lee: looks like this is the setup for your test only. In this case, better to have it inside another object. This way mocha will only run it before your tests. With the current setup, mocha will run this before ALL the tests

integration/sync/test_collision-handlers.js
```diff
@@ -0,0 +1,115 @@
...
+      mongodb = db;
+      mongodb.dropCollection(DATASETID + '_collision');
+      defaultHandlers = defaultDataHandlersModule(mongodb);
+      dataHandlers = dataHandlersModule({
+        defaultHandlers: defaultHandlers
+      });
+      return done(err);
+    });
+  },
+  'afterEach': function(done) {
```
wei-lee: similarly to the above, put this inside another object. Otherwise this will run before every test

integration/sync/test_collision-handlers.js
```diff
@@ -0,0 +1,115 @@
...
+      dataHandlers.removeCollision(DATASETID, res.uid, {}, function(err) {
+        assert.ok(!err);
+        dataHandlers.listCollisions(DATASETID, {}, function(err) {
+          assert.ok(!err);
+          assert.ok(customHandler.calledOnce);
+          return done();
+        });
+      });
+    });
+  }
```
wei-lee: can we also add some tests to cover the `listCollisions` and `removeCollision` in 'sync/index.js' file as well? Mainly because that is the API we exposed and there should be some tests to cover those.

aidenkeating:
@wei-lee Made the changes suggested.

Also fixed an issue with the collision being removed via the UID instead of the hash.

PR #74, opened 2017-03-02
Use ObjectID when referencing _id in MongoDB

aidenkeating:
Currently we are trying to remove and update records in mongo using
`fields: { _id: <recordId> }`. This won't work, the ID needs to be
converted into a mongodb.ObjectID object first.

We didn't see this before because it was hidden behind fh-db.


aidenkeating:
@david-martin Would you mind taking a look? I'm fairly sure this is correct

wei-lee:
This is now superseded by PR #78. Closing this.

PR #75, opened 2017-03-02
[WIP] Fix invalid bulk operation from MongoDB when adding empty array

aidenkeating:
Currently we allow for empty arrays to be added to the queue,
apparently MongoDB is a bit sensitive about that and throws an error.

This was discovered in the tests and the solution was discovered here:
https://github.com/toymachiner62/node-mongo-seeds/issues/6

A patch to `mongodb-queue` is also on the way.

### DIFF
--- a/lib/sync/mongodbQueue.js
+++ b/lib/sync/mongodbQueue.js
@@ -87,8 +87,11 @@ MongodbQueue.prototype.getName = function() {
 };
 
 MongodbQueue.prototype.addMany = function(messages, cb) {
+  if (messages.length === 0) {
+    return cb(new Error('Message list length must be greater than 0'));
+  }
   var self = this;
-  self.add(messages, cb); //it supports an array of messages since v2.2.0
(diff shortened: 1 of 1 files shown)


aidenkeating:
Just a little test script to check the empty array is causing this:

```javascript
var mongodb = require('mongodb');
mongodb.connect('mongodb://localhost:27017/test123', function(err, db) {
    db.collection('1234').insertMany([], function() {});
});
```

lib/sync/mongodbQueue.js
```diff
@@ -87,8 +87,10 @@ MongodbQueue.prototype.getName = function() {
 };
 
 MongodbQueue.prototype.addMany = function(messages, cb) {
-  var self = this;
-  self.add(messages, cb); //it supports an array of messages since v2.2.0
+  if (messages.length > 0) {
+    var self = this;
+    self.add(messages, cb); //it supports an array of messages since v2.2.0
+  }
```
wei-lee: we need to invoke the callback if the message array is empty

PR #77, opened 2017-03-03
add implementation for stop & stopAll. Update tests

wei-lee:
This PR will add the implementation for `sync.stop` and `sync.stopAll`. They are mainly used in tests.

ping @david-martin @aidenkeating @danbev for review

### DIFF
--- a/lib/sync/index.js
+++ b/lib/sync/index.js
@@ -285,7 +285,6 @@ module.exports = {
     stop: stop,
     stopAll: stopAll,
     connect: connect,
-    start: start, //TODO: looks like we don't need to expose this one anymore? init can be used as start
     toJSON: toJSON,
     setLogLevel: setLogLevel,
     globalHandleList: function() {},
--- a/lib/sync/index.js
+++ b/lib/sync/index.js
@@ -78,27 +78,59 @@ function invoke(dataset_id, params, callback) {
 
 // Initialise cloud data sync service for specified dataset.
 function init(dataset_id, options, cb) {
-  syncUtil.doLog(dataset_id, 'info', 'init');
-  datasets.init(dataset_id, options);
-  return cb && cb(null, {});
+  syncUtil.doLog(dataset_id, 'info', 'init sync for dataset');
+  start(function(err){
+    if (err) {
+      return cb(err);
+    }
+    datasets.init(dataset_id, options);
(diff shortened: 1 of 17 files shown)


lib/sync/index.js
```diff
@@ -253,7 +285,7 @@ module.exports = {
     stop: stop,
     stopAll: stopAll,
     connect: connect,
-    start: start,
+    start: start, //TODO: looks like we don't need to expose this one anymore? init can be used as start
```
david-martin: @wei-lee  Agreed. Given that a call into `sync` or `syncRecords` will start the relevant sync things if not already done, this can be removed.

lib/sync/worker.js
```diff
@@ -80,4 +83,12 @@ QueueWorker.prototype.work = function() {
   });
 };
 
+/**
+ * Stop the queue worker
+ */
+QueueWorker.prototype.stop = function(cb){
```
david-martin: as this is synchronous, can leave out a callback. No need to return anything either. If function finishes, the caller knows it was marked as stopped.

lib/sync/index.js
```diff
@@ -78,27 +78,59 @@ function invoke(dataset_id, params, callback) {
...
+    mongoDbClient = null;
+    redisClient = null;
+    metricsClient = null;
+    ackQueue = null;
+    pendingQueue = null;
+    syncQueue = null;
+    ackWorker = null;
+    pendingWorker = null;
+    syncWorker = null;
+    syncStarted = false;
```
david-martin: Given that these are all scoped to this module, and this module (index.js) has multiple functions:
* exposing handler functions for fh-mbaas-api to use
* exposing the API
* managing the starting/stopping of a Sync Server
* managing the various parts that make up a Sync Server (queues, locks, workers, etc..)
Perhaps a good change is moving the Sync Server parts out of index.js into its own module (following the prototype approach similar to other sync files).
Interested in your thoughts as a follow up change (not in this PR)

wei-lee: yeah, good point. I see no reason why we can't do that. Will make index.js a little bit cleaner as well.

lib/sync/worker.js
wei-lee: although the implementation is synchronous, I would like to give it a callback style interface, just incase this could change in the future.

PR #80, opened 2017-03-06
add sync event hooks

evanshortiss:
NOTE: Will need to be merged #53 first since this is branched off of it and therefore the diff is a little heavy right now.

This PR would allow developers to hook into sync events and define behaviours based on this. 
Here's an example:

```js
const sync = require('fh-mbaas-api').sync;
const stats = require('lib/stats');
const activities = require('lib/models/activities');

// note event args is hash that matches doList params minus the callback
sync.events.on('list-success', function (dataset_id, args, time) {
  // Might be a fire and forget function that tracks operation call counts
  stats.increment('sync-list', dataset_id);

  console.log(args) // prints sync call args { params: params, meta_data: meta_data } 
})

// note event args is a hash that matches doUpdate params minus the callback
sync.events.on('update-success', function (dataset_id, args, time) {
  console.log('update took ${time} milliseconds');
  stats.increment('sync-update', dataset_id);

  // Perhaps we need to track updates for a home page event list for admins
  activities.record(`User ${args.meta_data.userId} updated record ${args.uid}`);
})
```

You could achieve this by writing custom handlers, but you might just want a very simple mechanism for recording events like this and writing custom handlers would be significantly more effort than hooking into an events

### DIFF
--- a/lib/sync-DataSetModel.js
+++ b/lib/sync-DataSetModel.js
@@ -18,7 +18,6 @@ function generateGlobalSetter (handlerName, inst) {
 }
 
 var self = {
-
   globalListHandler: undefined,
   globalCreateHandler: undefined,
   globalReadHandler: undefined,
@@ -39,7 +38,7 @@ var self = {
     "clientSyncTimeout": 15,
     "logLevel": "info"
   },
-
+  events: undefined,
   datasets: {},
   datasetClientRecords: {},
   deletedDatasets: {},
@@ -51,6 +50,14 @@ var self = {
     defaultDataHandler.setFHDB(db);
   },
 
+
+  /**
+   * Pass an event emitter instance for this module to use
+   * @param {EventEmitter} ee
+   */
+  setEventEmitter: function (ee) {
+    self.events = ee;
+  },
 
   /**
    * Returns the hash fn for hashing records for a given dataset
@@ -107,6 +114,9 @@ var self = {
   doListHandler: function (dataset_id, query_params, cb, meta_data) {
     self.getDataset(dataset_id, function (err, dataset) {
       if (err) return cb(err);
+
+      self.events.emit('list', dataset_id, query_params, meta_data);
+
       if (dataset.listHandler) {
         dataset.listHandler(dataset_id, query_params, cb, meta_data);
       }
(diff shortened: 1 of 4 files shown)


evanshortiss:
@wei-lee let me know your thoughts if you get a chance

lib/sync-DataSetModel.js
```diff
@@ -107,66 +133,147 @@ var self = {
...
+      //     uid: uid,
+      //     pre: pre,
+      //     post: post,
+      //     meta_data: meta_data
+      //   },
+      //   cb
+      // );
+
       if (dataset.collisionHandler) {
         dataset.collisionHandler(dataset_id, hash, timestamp, uid, pre, post, meta_data);
```
wei-lee: I think we should change the `collisionHandler` here to accept a callback function if it's defined to keep the API consistent.

@evanshortiss WDYT?

wei-lee:
@evanshortiss I think your PR is good, with just one comment around the collisonDataHandler. I think we can try make it accept callback function as well, and then we can keep the API consistent, but also backward compatible.

I am also thinking should we add more events across the sync server. Do you think it will be something that could be useful?

evanshortiss:
@wei-lee I agree, keeping the collision handler consistent would be ideal. Happy to look into it in this PR.

I think more events could definitely be useful. If you have some in mind I would be happy to add here also.

david-martin:
This PR looks related to https://github.com/feedhenry/fh-mbaas-api/pull/101.
@evanshortiss If a 'sync-ready' event is a type of event you would see being emitted via sync.events, Maybe we can merge this PR and just call sync.events.emit('sync-ready') instead of the changes in https://github.com/feedhenry/fh-mbaas-api/pull/101?

david-martin:
@evanshortiss On further thought, I'm hesitant to merge these changes for the existing sync server code, as we'll be moving towards the newer sync implementation on this branch https://github.com/feedhenry/fh-mbaas-api/tree/FH-3123-sync-horizontal-scaling

@wei-lee had a good comment about namespacing events too. Here it is https://github.com/feedhenry/fh-mbaas-api/pull/101#pullrequestreview-27822438

Maybe we can move the conversation there?

david-martin:
Closing as affected files are all scheduled for removal in https://issues.jboss.org/browse/FH-3242, superseded by the implementation in https://github.com/feedhenry/fh-mbaas-api/tree/FH-3123-sync-horizontal-scaling.
These changes can be re-applied if needed based on that branch and #101, or wait until they are both back on master.

PR #81, opened 2017-03-06
RHMAP-12743 - databrowser export doesn't work if database not upgraded

rachael-oregan:
## Motivation
Exporting data from databrowser doesn't work if the database is not upgraded, regardless of the selected format. The following error is thrown: 
{"message":"write after end","code":500}

The request to export was ended before the action was taken.

## Result
Remove the `req.end()`. Not sure why this was added.

Note: exporting of bson format doesn't work https://issues.jboss.org/browse/RHMAP-12745

## Jira 
https://issues.jboss.org/browse/RHMAP-12743

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -190,11 +190,12 @@ function net_db(params, callback) {
       logger.warning('Problem contacting DITCH server: ' + err.message + "\n" + util.inspect(err.stack));
       return callback(err);
     });
-    req.end();
+
+    req.on('end', function () {
+      return callback(null, {stream: req})
+    });
 
     return callback(null, {stream: req});
   }
   return;
-
-
 }
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,6 +1,6 @@
 {
   "name": "fh-mbaas-api",
-  "version": "6.1.5",
+  "version": "6.1.6",
   "dependencies": {
     "async": {
       "version": "0.2.9",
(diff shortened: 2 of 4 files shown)


lib/db.js
```diff
@@ -190,7 +190,6 @@ function net_db(params, callback) {
       logger.warning('Problem contacting DITCH server: ' + err.message + "\n" + util.inspect(err.stack));
       return callback(err);
     });
-    req.end();
```
pb82: Looks like we lost all the git history of this component with open sourcing it. I could not figure out when and why this was added. Request objects should fire the `end` event if no more data is provided. Maybe instead of removing it you can try this:

```javascript
req.on('end', function () {
    callback(null, {stream: req});
});
```

(Return only when the stream has finished)

trepel: Yes, I also tried to find the commit responsible for this but was not successful. It should be possible if you clone the old fheng/fh-mbaas-api repo but I haven't found easy way how to identify the commit.

pb82: @trepel Right, looks like it was introduced with this commit: https://github.com/fheng/fh-mbaas-api/commit/3170fe119fa7abc981fe3d10c7e05afb9d80df42#diff-be228df81b99a384aa50ad50bde7fca0R114

No real explanation given though.

trepel: @pb82 @rachael-oregan I must admit I have little-to-zero understanding of the code but having the line `callback(null, {stream: req});` twice (line 195 and line198) in codebase looks weird to me. 
When I removed the line 198 the export didn't work. The error was:
`You cannot pipe after data has been emitted from the response.`

If you are sure the code is ok as is, feel free to ignore this comment :)

PR #84, opened 2017-03-07
Fh 3194 clean queues

wei-lee:
This PR adds new functions to automatically remove acked messages from the queue. It allows some messages to be left in the queue so they can be used as an audit log. By default, it will keep the messages in the last 24h and remove all the other acked messages. It will run every hour.


integration/sync/test_mongodbQueue.js
```diff
@@ -0,0 +1,86 @@
...
+    },
+
+    'after': function(done) {
+      if (queue) {
+        queue.stopPruneJob();
+      }
+      done();
+    },
+
+    'test mongo queue add, get, search and prune': function(done) {
```
wei-lee: @grdryn do you know where it fails at?

grdryn: @wei-lee I'm not sure. It's possible that instead of it being a problem with this test, that it's a problem with how `npm test` runs the tests: it could be that the pre-test script that configures the container with MongoDB doesn't wait for it to be ready, and then the test can't connect properly, or something?

PR #87, opened 2017-03-08
Minor cleanup in lib/sync/util before FH-3220

grdryn:
@david-martin @wei-lee @aidenkeating would any of you like to review?

These changes basically come from looking at FH-3220, and trying to decide what to do with the logging line, I figured that it could maybe be removed (I can't find any inputs that `JSON.stringify` throws for), and then just cleaned up some other things in the process.

Does fh-mbaas-api just get used in NodeJS, or does it also get used in the browser. If it gets used in the browser, will changing the function expressions to function declarations here cause any issues? I'm not as familiar with JS in browsers. If we're not sure, I could change them to named function expressions instead. Basically I was just trying to not have anonymous functions.

### DIFF
--- a/lib/sync/util.js
+++ b/lib/sync/util.js
@@ -40,15 +40,9 @@ var sortObject = function (object) {
 
 var sortedStringify = function (obj) {
   var str = '';
-  try {
-    if (obj) {
-      str = JSON.stringify(sortObject(obj));
-    }
-  } catch (e) {
-    doLog(SYNC_LOGGER, 'error', 'Error stringifying sorted object:' + e);
-    throw e;
+  if (obj) {
+    str = JSON.stringify(sortObject(obj));
   }
-
   return str;
 }
 
--- a/lib/sync/util.js
+++ b/lib/sync/util.js
@@ -39,11 +39,7 @@ var sortObject = function (object) {
 
 
 var sortedStringify = function (obj) {
-  var str = '';
-  if (obj) {
-    str = JSON.stringify(sortObject(obj));
-  }
-  return str;
+  return obj ? JSON.stringify(sortObject(obj)) : '';
 }
 
 var setLogger = function (dataset_id, options) {
--- a/lib/sync/util.js
+++ b/lib/sync/util.js
@@ -7,7 +7,7 @@ var _ = require('underscore');
 var SYNC_LOGGER = 'SYNC';
 var loggers = {};
 
-var generateHash = function (plainText) {
+function generateHash(plainText) {
   var hash;
   if (plainText) {
     if ('string' !== typeof plainText) {
@@ -20,7 +20,7 @@ var generateHash = function (plainText) {
   return hash;
 }
 
-var sortObject = function (object) {
+function sortObject(object) {
   if (typeof object !== "object" || object === null) {
     return object;
   }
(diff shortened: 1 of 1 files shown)


paolobueno:
`JSON.stringify` will throw for circular references: `let a = {}; a.x = a; JSON.stringify(a) // TypeError: Converting circular structure to JSON`

lib/sync/util.js
```diff
@@ -7,7 +7,7 @@ var _ = require('underscore');
 var SYNC_LOGGER = 'SYNC';
 var loggers = {};
 
-var generateHash = function (plainText) {
+function generateHash(plainText) {
```
wei-lee: interesting to see that you changed from function expression to function declaration. Just curious any reason you doing this?

There is a subtle difference between the 2, and many suggests to use function expressions if you can.

wei-lee:
As @paolobueno pointed out, JSON.stringify can throw exceptions.

lib/sync/util.js
grdryn: @wei-lee what I was trying to do was not have anonymous functions. I guess that we could have named function expressions instead of function declarations. E.g. `var generateHash = function generateHash(plainText) {`.

Since the main purpose of this PR was to get rid of the try/catch, which actually seems necessary now, I'll just close the PR instead. :)

grdryn:
@paolobueno @wei-lee thanks for the feedback on `JSON.stringify`, I'll close the PR based on that, since the other changes are not really necessary. 👍

david-martin:
@grdryn To answer 1 of your questions, `fh-mbaas-api` is not intented to be used in the browser.

grdryn:
@david-martin thanks, that's useful to know. :)

lib/sync/util.js
grdryn: @wei-lee thinking about this more, do we have any guidelines (or are there any existing 3rd party guidelines that we can refer to), on when to use one over the other? AIUI the main difference is scoping, so either should be fine in this case.

There are other places in this module where function declarations are used in cases like this (such as in `lib/sync/index.js`). I think it would be good to have a consistent way of knowing which to use where (like "use X instead of Y, except in these specific circumstances").

Unless we have such a thing already, and I'm just not aware of it! :)

lib/sync/util.js
```diff
@@ -38,21 +38,11 @@ var sortObject = function (object) {
...
 
 
-var sortedStringify = function (obj) {
-  var str = '';
-  try {
-    if (obj) {
-      str = JSON.stringify(sortObject(obj));
-    }
-  } catch (e) {
-    doLog(SYNC_LOGGER, 'error', 'Error stringifying sorted object:' + e);
```
grdryn: @paolobueno @wei-lee Thinking about the `JSON.stringify` example here again (I just happened to be passing the same code again in my editor), how useful is the try/catch really, given that we're just going to re-throw the error anyway?

AFAICT, all we get from keeping is an additional error log message saying `Error stringifying sorted object: [object Object]`. At least I think that's how it will come out, given that it *is* an object and not null or undefined or another type -- none of those seem to cause an error to be thrown. Are there other inputs for which it would throw an error?

lib/sync/util.js
wei-lee: @grdryn I don't think we have such a thing. But personally I do use both, and I am not sure if we need to specify which one to use. But instead, we should set a rule that always declare the function BEFORE invoking it. This will avoid the "hoist" problem, and makes it easier to understand as well.

PR #88, opened 2017-03-08
🎨 Try to make logging a little more consistent

grdryn:
JIRA: https://issues.jboss.org/browse/FH-3220

Motivation:

See the conversation in the JIRA issue for more details, but basically we want to try to achieve some sort of consistency in the logging from the sync framework.

Modification:

We attempt to follow the suggestions in the bunyan level suggestions section:

https://github.com/trentm/node-bunyan#level-suggestions

Since this is a library rather than an app, almost all logs are turned down to "debug" level (bunyan suggest "trace", but not all loggers have a "trace" level, such as winston, which is what is currently being used).

Result:

* Most log entries are debug level
* Some log entries are error level, when they add context to an error that occurs, and it's difficult to pass that information up to a caller.

Additional Notes:

It's not always clear whether we're passing an `Error` object or a `String` as the *err* first param to a callback. In fact, there's at least one function here in the sync code which passes an `Error` object in one case, and a `String` in another case.

That may be something that's difficult to change now. If we could though, and we choose to pass an `Error` object always, we could consider using the [verror](https://www.npmjs.com/package/verror) module to wrap errors with context as they bubble up to the app, rather than having to log at error level at the points we do.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,6 @@
-#Overview
+# fh-mbaas-api
+
+fh-mbaas-api provides FeedHenry MBaaS APIs for Node.js cloud apps.
 
 [![npm package](https://nodei.co/npm/fh-mbaas-api.png?downloads=true&downloadRank=true&stars=true)](https://nodei.co/npm/fh-mbaas-api/)
 
@@ -6,8 +8,6 @@
 [![Dependency Status](https://img.shields.io/david/feedhenry/fh-mbaas-api.svg?style=flat-square)](https://david-dm.org/feedhenry/fh-mbaas-api)
 [![Known Vulnerabilities](https://snyk.io/test/npm/fh-mbaas-api/badge.svg?style=flat-square)](https://snyk.io/test/npm/fh-mbaas-api)
 
-
-fh-mbaas-api provides FeedHenry MBaaS APIs to Node.js cloud apps.
 
 |                 | Project Info  |
 | --------------- | ------------- |
@@ -18,7 +18,7 @@ fh-mbaas-api provides FeedHenry MBaaS APIs to Node.js cloud apps.
 | Mailing list:   | [feedhenry-dev](https://www.redhat.com/archives/feedhenry-dev/) ([subscribe](https://www.redhat.com/mailman/listinfo/feedhenry-dev))  |
 | IRC:            | [#feedhenry](https://webchat.freenode.net/?channels=feedhenry) channel in the [freenode](http://freenode.net/) network.  |
 
-#Usage
+## Usage
 fh-mbaas-api is included as standard with your cloud app code.
 
 For custom apps, add the module via npm by running the following for the root of your app
@@ -29,18 +29,18 @@ npm install --save fh-mbaas-api
 
 This will install the latest version of fh-mbaas-api and save the installed version in your package.json
 
-#Documentation
+## Documentation
 Documentation for the $fh cloud API is maintained at the [FeedHenry API Docs.](http://docs.feedhenry.com/v3/api/cloud_api.html)
 
-#Deprecated
-Legacy Rhino functions have been deprecated. These are listed below - with their replacements **in bold**. All replacements listed but '$fh.web' have drop-in replacements available.  
+## Deprecated
+Legacy Rhino functions have been deprecated. These are listed below - with their replacements **in bold**. All replacements listed but '$fh.web' have drop-in replacements available.
 
 * $fh.web -> **[request](https://github.com/mikeal/request)**
 * $fh.log -> **console.log**
 * $fh.parse -> **JSON.parse**
 * $fh.stringify  **JSON.stringify**
 
-#Tests
+## Tests
 In order to run the tests, please make sure you have [Docker](https://www.docker.com/) installed.
 
 Before running tests do:
@@ -54,9 +54,9 @@ Then to run the tests use ```npm test```
 
 On Windows, use ```npm run testwindows```
 
-# Caveats
+## Caveats
 
-## Two sync loops per sync frequency
+### Two sync loops per sync frequency
 Two sync loops may be invoked per sync frequency if the server-side sync frequency
 differs from the client-side frequency.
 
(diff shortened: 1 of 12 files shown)


lib/sync/api-syncRecords.js
```diff
@@ -183,8 +183,9 @@ function syncRecords(datasetId, params, cb) {
           return callback(err);
         }
         if (!datasetClientJson) {
```
grdryn: This is the part that I mentioned in the commit message, where it returns a different type as the first parameter to the callback in different situations.

lib/sync/sync-processor.js
```diff
@@ -52,7 +52,8 @@ function recordProcessTime(startTime, success) {
 function markDatasetClientAsCompleted(datasetClientId, startTime, callback) {
   syncStorage.updateDatasetClient(datasetClientId, {syncLoopEnd: Date.now(), syncCompleted: true, syncScheduled: null}, function(err){
     if (err) {
-      syncUtil.doLog(datasetId, "error", "Error when update dataset client with id " + datasetClientId + ". error = " + util.inspect(err));
+      // TODO: datasetId here isn't defined, is it?
```
grdryn: @wei-lee syntax highlighting in my editor showed me that there's no `datasetId` and I think I agree. I'm not sure what the first param should be here, `syncUtil.SYNC_LOGGER` or `datasetClientId` or something?

grdryn:
Retest this please.

lib/sync/sync-processor.js
aidenkeating: @grdryn I think here, if we have no access to the datasetId, we should be using `syncUtil.SYNC_LOGGER`. The first argument is meant to define the dataset to log for, any other place where there is no access to a dataset id seems to use `syncUtil.SYNC_LOGGER`.

If we could change this function to accept a `DatasetClient` object we could use the `getId()` and `getDatasetId()` methods to make this complete. 

I haven't looked much into it but I'm not sure that this is possible though without calling `doReadDatasetClient` which would have to call out to MongoDB, which is probably a bit overkill just to provide a log.

wei-lee: yeah, we should just use the `syncUtil.SYNC_LOGGER` here

wei-lee:
@grdryn I think we should stick to error objects rather than string messages. But we probably don't have to do it now, might be better create another ticket for it and we can progress it later.

grdryn:
@wei-lee sure. I've created [FH-3249](https://issues.jboss.org/browse/FH-3249) for that just now.

lib/sync/sync-processor.js
grdryn: Ack, I'll change this.

lib/sync/index.js
```diff
@@ -46,12 +46,13 @@ var syncLock = null;
 var interceptors = null;
 var apiSync = null;
 var apiSyncRecords = null;
+var syncScheduler = null;
```
grdryn: This gets initialized in `.start`, and accessed in `.stopAll`, so it should be declared here for both.

grdryn:
I've updated to replace the discussed undefined `datasetId`, and also to add a variable declaration in `index.js` without which I think it could be out of scope.

grdryn:
@david-martin I've added a note to the README about logging in the Sync API here.

I think this is ready for review again if anyone would like to take a look! :)

cc @wei-lee @wojta

david-martin:
@grdryn :+1: on section in the README. (fyi you'll need a rebase)

grdryn:
Rebased, I'll merge in a while if the jobs pass, and if nobody has any objections.

README.md
```diff
@@ -54,19 +54,28 @@ Then to run the tests use ```npm test```
...
 
-The `syncFrequency` value of the dataset on the server should be set to the 
+The `syncFrequency` value of the dataset on the server should be set to the
 `sync_frequency` value of the corresponding dataset on the client to avoid this.
 
 For example:
   * `sync_frequency` on the client-side dataset is also set to 120 seconds.
   * `syncFrequency` on the server-side dataset is set to 120 seconds.
+
+## Sync API logging
```
evanshortiss: @grdryn @wei-lee I like this idea of having a more standard approach to logging. Would you consider having a standard logger across the entire fh-mbaas-api? It seems sync does not share the common logger found [here](https://github.com/feedhenry/fh-mbaas-api/blob/master/lib/consolelogger.js). If it did then enabling/disabling SDK logs would be easy to approach.

I find the `debug` module is a nice to use logger in libraries since it can easily be disabled and enabled like so:

```
export DEBUG=name-of-logger
```

So for the mbaas-api we could do things like:

```
# just enable sync and service verbose logs
export DEBUG=fh-mbaas-api:sync,fh-mbaas-api:service

# enable all fh-mbaas-api logs
export DEBUG=*
```

Then in code you can have:

```
// sync.js

// you could wrap debug to add custom log formatting...
var log = require('debug')('fh-mbaas-api:sync');
```

For some apps built by consulting we use [fh-bunyan](https://www.npmjs.com/package/fh-bunyan) too since it auto configures itself based on the `FH_ENV`. When running locally (FH_USE_LOCAL_DB is set) it will use `trace` level. It can also be configured by setting an `BUNYAN_FH_LEVEL` env var.

Just some of my 2 cents. Nice work on the recent sync stuff! 👍

wojta: Yes, we have separate issue for that. https://issues.jboss.org/browse/FH-3219. Thanks for suggestions.

grdryn: @evanshortiss like @wojta said, we've got another issue in progress to make additional changes to logging, specifically to allow a logger to be passed in, rather than the current default. I do like the sound of the idea of using the debug module as a default, if we have a default. That might be done in a separate issue again after FH-3219, I'm not sure.

evanshortiss: @wojta @grdryn good to hear it's on the radar. Passing in a logger is definitely a neat feature if feasible! Thanks for the JIRA link

PR #89, opened 2017-03-09
 🐛 Fix 2 errors discovered by the load tests as described in https://…

wei-lee:
…issues.jboss.org/browse/FH-3246

### DIFF
--- a/lib/sync/api-syncRecords.js
+++ b/lib/sync/api-syncRecords.js
@@ -19,7 +19,7 @@ function listLocalDatasetClientData(datasetClient, cb) {
     }
     //no one sync loop has completed yet, return null
     if (!datasetClientsWithRecords.syncCompleted) {
-      return cb();
+      return cb(null, null);
     } else {
       return cb(null, datasetClientsWithRecords);
     }
--- a/lib/sync/default-dataHandlers.js
+++ b/lib/sync/default-dataHandlers.js
@@ -4,6 +4,16 @@ var ObjectID = require('mongodb').ObjectID;
 
 var mongo;
 
+function convertToObjectId(datasetId, originalId) {
+  var newObjectId = originalId;
+  if (ObjectID.isValid(originalId)) {
+    newObjectId = ObjectID(originalId);
+  } else {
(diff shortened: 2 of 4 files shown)


wei-lee:

lib/sync/default-dataHandlers.js
```diff
@@ -43,7 +53,7 @@ module.exports = function (db) {
                      'doRead : ' + dataset_id +
                       ' :: ' + uid +
                       ' :: meta=' + util.inspect(meta_data));
-      mongo.collection(dataset_id).findOne({"_id" : ObjectID(uid)}, function(err, found){
+      mongo.collection(dataset_id).findOne({"_id" : convertToObjectId(dataset_id, uid)}, function(err, found){
```
aidenkeating: Just a question really. I am wondering whether we are handling when `found` here is `null` or `undefined` in sync? 

In the case where a non-ObjectID is returned by `convertToObjectId` both `err` and `found` will be `undefined` for mongodb.findOne, I think.

Do we handle this somewhere or will we expect the handler implementation to handle it? Or, does it even need to be handled?

lib/sync/api-syncRecords.js
```diff
@@ -19,7 +19,7 @@ function listLocalDatasetClientData(datasetClient, cb) {
     }
     //no one sync loop has completed yet, return null
     if (!datasetClientsWithRecords.syncCompleted) {
-      return cb();
+      return cb(null, null);
```
aidenkeating: I wouldn't have thought this would make much of a difference, does it?

wei-lee: It actually does. See the test I have added in this PR. It will fix the other error we saw in the ticket. The problem is that this is being used as part of `async.waterfall`, and it nothing is returned (`undefined`), then the `async` module will not pass it to the next function. But if it's `null` value, `async` will still pass it as the first parameter to the next function.

lib/sync/default-dataHandlers.js
wei-lee: I think we just need to make sure `found` is not null if we are doing anything with it. Otherwise we can just pass it to the caller and let the caller decide what to do.

PR #91, opened 2017-03-09
⚡ Improve mongodb query performance by adding the missing indexes

wei-lee:
During the load test, we noticed that some mongodb operations are very slow. This PR will add indexes on some of the collections to see if it will improve the performance.


david-martin:
@wei-lee :+1: 
Before merging, can we try out these changes first to prevent reverting/rebased the change after?

wei-lee:
@david-martin I have run the tests locally already and it works fine.

PR #92, opened 2017-03-10
 🐛  do not delete the datasetclients if they are inactive

wei-lee:
This PR will not remove the dataset clients immediately if they are inactive. We will not sync them but leave them in the db.

There will be another PR to add the job that will clean then up.

### DIFF
--- a/lib/sync/sync-scheduler.js
+++ b/lib/sync/sync-scheduler.js
@@ -49,9 +49,9 @@ SyncScheduler.prototype.checkDatasetsForSyncing = function(cb) {
       var datasetId = datasetClient.getDatasetId();
       if (datasetClient.shouldDeactiveSync()) {
         //if the a datasetclient is not active, we don't need to sync it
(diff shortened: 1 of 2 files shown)


lib/sync/sync-scheduler.js
```diff
@@ -43,20 +43,19 @@ SyncScheduler.prototype.checkDatasetsForSyncing = function(cb) {
...
     var datasetClientsToSync = [];
-    var datasetClientsToRemove = [];
 
     _.each(datasetClients, function(datasetClientJson){
       var datasetClient = DatasetClient.fromJSON(datasetClientJson);
       var datasetId = datasetClient.getDatasetId();
-      if (datasetClient.shouldSync()) {
+      if (datasetClient.shouldDeactiveSync()) {
+        //if the a datasetclient is not active, we don't need to sync it
+        syncUtil.doLog(datasetId, 'info', 'Deactivating sync for client ' + datasetId + '. No client instances have accessed in ' + datasetClient.getConfig().clientSyncTimeout * 1000 + 'ms');
```
david-martin: Can we drop this log level to `debug`?
See https://issues.jboss.org/browse/FH-3220 for more info.
Hoping to have something in the README with the reasons why we'll only log at `debug`

wei-lee:
comment addressed. merging.

PR #93, opened 2017-03-10
✨  add a new backgroup task that will cleanup inactive dataset client…

wei-lee:
…s periodically

### DIFF
--- a/lib/sync/datasetClientsCleaner.js
+++ b/lib/sync/datasetClientsCleaner.js
@@ -1,0 +1,78 @@
+var util = require('util');
+var parseDuration = require('parse-duration');
+var async = require('async');
+var _ = require('underscore');
+var DatasetClient = require('./DatasetClient');
+var syncUtil = require('./util');
+
+var storage;
+
+/**
+ * Constructor for DatasetClientsCleaner.
+ * 
+ * A DatasetClientsCleaner will check the status of all the dataset clients in the db, and if any of them hasn't been accessed for the past <retentionPeriod> time, it will remove them from the db to save space.
+ * By default, it will remove dataset clients that haven't been active for 1 day and it will run the check every hour.
+ * Developer can change those parameters via the configuration.
+ * 
+ * @param {Object} opts configuration options for the DatasetClientsCleaner. 
+ * @param {String} retentionPeriod the maximum time the dataset client can be kept in db after it's last access tim. Support format: https://www.npmjs.com/package/parse-duration. Default is '24h'.
+ * @param {String} checkFrequency how often the checks should be running. Default is '1h'.
+ */
+function DatasetClientsCleaner(opts) {
+  this.retentionPeriod = opts.retentionPeriod || '24h';
+  this.checkFrequency = opts.checkFrequency || '1h';
+  this.job;
+}
+
+function cleanDatasetClients(retentionPeriod) {
+  async.waterfall([
+    async.apply(storage.listDatasetClients),
+    function filterDatasetClients(datasetClients, callback) {
+      var datasetClientsToRemove = _.filter(datasetClients, function(datasetClientJson){
+        var datasetClient = DatasetClient.fromJSON(datasetClientJson);
+        return datasetClient.shouldBeRemoved(parseDuration(retentionPeriod));
+      });
+      return callback(null, datasetClientsToRemove);
+    },
+    async.apply(storage.removeDatasetClients)
+  ], function(err, deletedDatasetClients){
+    if (err) {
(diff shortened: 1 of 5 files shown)


lib/sync/datasetClientsCleaner.js
```diff
@@ -0,0 +1,78 @@
+var util = require('util');
+var parseDuration = require('parse-duration');
+var async = require('async');
+var _ = require('underscore');
+var DatasetClient = require('./DatasetClient');
+var syncUtil = require('./util');
+
+var storage;
```
grdryn: I'm not that familiar with the prototypical OO stuff in JS, so this probably won't make a lot of sense: Would this make sense as an instance field? I guess not, since `cleanDatasetClients` uses it, and that's a function rather than a method (would that make sense to be a method, or maybe pass `storage` as a parameter to the function?)?

test/sync/test_datasetClientsCleaner.js
```diff
@@ -0,0 +1,64 @@
...
+        assert.ok(syncStorage.listDatasetClients.called);
+        assert.ok(syncStorage.removeDatasetClients.called);
+        var datasetClientsToRemove = syncStorage.removeDatasetClients.args[0][0];
+        assert.equal(datasetClientsToRemove.length, 1);
+        assert.equal(datasetClientsToRemove[0].queryParams.user, '1');
+        done();
+      }, 120);
+    },
+    'test datasetClientCleaner with no dataset clients': function(done) {
+      var datasetClients = [];
```
grdryn: Could `datasetClients` (from `syncStorage.listDatasetClients()`) ever be null or undefined or something that's not an array? If so, what would happen? Would a test for such a case be useful?

lib/sync/datasetClientsCleaner.js
wei-lee: I think either way will work. The thing here is that I want `cleanDatasetClients` to be private. If I add it to the instance, it will become public.

test/sync/test_datasetClientsCleaner.js
wei-lee: Good question 😄 . I double checked `syncStorage.listDatasetClients` and it should not return null. If the value is null, then there must be an error some where. In that case, `async.waterfall` should just invoke the callback and skip the rest of the functions in the array.

lib/sync/datasetClientsCleaner.js
```diff
@@ -0,0 +1,78 @@
...
+  this.job;
+}
+
+function cleanDatasetClients(retentionPeriod) {
+  async.waterfall([
+    async.apply(storage.listDatasetClients),
+    function filterDatasetClients(datasetClients, callback) {
+      var datasetClientsToRemove = _.filter(datasetClients, function(datasetClientJson){
+        var datasetClient = DatasetClient.fromJSON(datasetClientJson);
+        return datasetClient.shouldBeRemoved(parseDuration(retentionPeriod));
```
david-martin: To avoid `parseDuration(retentionPeriod)` being called multiple times (once per dataset client) it could probably be moved outside the `_.filter` or even the `waterfall` at the top. I'm assuming the result of that function is deterministic?

wei-lee: yes, we could do that.

wei-lee:
@david-martin Yes, those scenarios you listed could happen, but I am not sure if they are really "problems". I want to keep things simple for now, and let us run a few tests and see if there is problem with it. If there is problem, we can then consider adding locks and restrict it to just one worker. Does it make sense?

david-martin:
@wei-lee I think the multiple workers/pods (or even multiple executions in 1 process) running the same queries doesn't qualify as a problem/bug, as I don't think it can break things.
I do think it could affect performance, but theres a number of factors at play before it would be significant.
If you want to defer such a change, that's OK.

lib/sync/datasetClientsCleaner.js
evanshortiss: Typically we've found avoiding the `prototype` usage when possible is nice. It avoids errors with `this` not being bound correctly. It looks like this constructor could do so?

```js
function DatasetClientsCleaner () {
  const instance = xtend({}, {
    job: null,
    retentionPeriod: opts.retentionPeriod || '24h',
    checkFrequency: opts.checkFrequency || '1h'
  });

  instance.start = function (immediately, callback) {
    if (immediately) {
      cleanDatasetClients(self.retentionPeriod);
    }
    if (!instance.job) {
      instance.job = setInterval(function() {
        cleanDatasetClients(instance.retentionPeriod);
      }, parseDuration(instance.checkFrequency));
    }
    return cb && cb();
  };

  return instance;
}
```

Obviously depends on how many of these will be created in memory. We also keep functions private similar to how you did so here @wei-lee 👍

PR #94, opened 2017-03-13
🐛 FH-3254 Add a null check for dataset client before processing

david-martin:
This change prevents a scenario where a datasetclient has been removed
from the db (via a cleanup task for inactive clients) yet there are
still pending items in the queue for that datasetclient.
In that scenario, it is best to mark the pending update as failed as we
don't have enough info to process it correctly.
The update failed notification will be triggered on the client whenever
they become active again and do a sync with the server

### DIFF
--- a/lib/sync/storage.js
+++ b/lib/sync/storage.js
@@ -136,6 +136,11 @@ function doUpdateDatasetClient(datasetClientId, fields, upsert, cb) {
       syncUtil.doLog(syncUtil.SYNC_LOGGER, 'error', 'Failed to update datasetClients due to error ' + util.inspect(err) + ' :: datasetClientId = ' + datasetClientId + ' :: fields = ' + util.inspect(fields));
       return cb(err);
     }
+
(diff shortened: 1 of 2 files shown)


lib/sync/storage.js
```diff
@@ -136,6 +136,11 @@ function doUpdateDatasetClient(datasetClientId, fields, upsert, cb) {
       syncUtil.doLog(syncUtil.SYNC_LOGGER, 'error', 'Failed to update datasetClients due to error ' + util.inspect(err) + ' :: datasetClientId = ' + datasetClientId + ' :: fields = ' + util.inspect(fields));
       return cb(err);
     }
+
+    if (result.value === null) {
+      return cb('DatasetClient not found for id ' + datasetClientId);
```
aidenkeating: Would it be better to provide an `Error` object here instead of a string?

david-martin: @aidenkeating Good suggestion. I'll make that change.

PR #95, opened 2017-03-14
 🐛 make sure the acknowledgements in the request body will not be ret…

wei-lee:
…urned again in the response

# JIRA

https://issues.jboss.org/browse/FH-3345


wei-lee:
ping @aidenkeating @david-martin

PR #96, opened 2017-03-15
FH-3121 Integrate sync-acceptance-testing into Travis build

david-martin:


.travis.yml
```diff
@@ -12,5 +13,20 @@ before_install:
...
+  # Include an updated job, with the acceptance tests. This way we're not
+  # running the 4.4.3 build twice with one build including acceptance tests.
+  - node_js: "4.4.3"
+    script:
+      - npm link
+      - git clone https://github.com/feedhenry/sync-acceptance-testing.git
+      - cd sync-acceptance-testing
+      - npm install
+      - npm link fh-mbaas-api
+      - npm test
```
wei-lee: looks like this will not run the unit & integration tests in fh-mbaas-api anymore. Is that right?

david-martin: Fixed to run unit & integration tests for node 0.10 & 4.

Acceptance tests are run for node 4 as another matrix job

PR #97, opened 2017-03-15
FH-3219 using debug module for logging

wojta:
Implementation of passing logger from api.js to sync code. Some logger abstraction was addded in [baselogger.js](lib/baselogger.js).

Still WIP, I'm doing some tweaking little bit.
Merge issue solving is in progress.


wojta:
I discovered that I based my code on master branch, but I should do it in feedhenry:FH-3123-sync-horizontal-scaling.

wojta:
Ok, obviously some tests are not yet passing. 
It's probably because of passing cfg parameter to the Sync API (It also contains logger in it). I think that it's not a breaking change, but tests probably needs to be changed a little bit.
But please @david-martin, @wei-lee please take a look, if it's not kind of dead end.

david-martin:
Note to any reviewers, this PR incroporates this PR https://github.com/feedhenry/fh-mbaas-api/pull/101 which is yet to merge to the base branch

lib/sync/index.js
```diff
@@ -98,43 +99,53 @@ function invoke(dataset_id, params, callback) {
...
-  sync: metricsModule.timeAsyncFunc(metricsModule.KEYS.SYNC_API_PROCESS_TIME, server.sync),
-  syncRecords: metricsModule.timeAsyncFunc(metricsModule.KEYS.SYNC_API_PROCESS_TIME, server.syncRecords),
-  listCollisions: metricsModule.timeAsyncFunc(metricsModule.KEYS.SYNC_API_PROCESS_TIME, server.api.listCollisions),
-  removeCollision: metricsModule.timeAsyncFunc(metricsModule.KEYS.SYNC_API_PROCESS_TIME, server.api.removeCollision),
-  setLogLevel: setLogLevel,
-  api: _.extend({}, server.api, {
-    connect: connect,
-    invoke: invoke,
-    toJSON: toJSON,
+module.exports = function(cfg) {
```
david-martin: @wojta  I don't think there's any changes here, just re-ordering of code blocks. Is that right?

david-martin:
@wojta The changes look reasonable.
Do you have a sample log output from sync before & after?
Also, will this change log output from other parts of fh-mbaas-api?

lib/sync/index.js
wojta: There is the change that allows me to set the logger. module.exports wasn't function before.

wojta:
@david-martin Before that it was like
`INFO this is a log message`, now it's 
`2017-03-17 12:07:24 INFO [tag] this is a log message`.  
[tag] is not used in old code, so it's not there. It's passing the same tests for the logger as the old code.

lib/consolelogger.js
```diff
@@ -1,62 +1,25 @@
...
+  if (tag !== "") {
+    tag = "[" + tag + "] ";
   }
+  console.log(this.getTS() + " " + baselogger.LEVEL_NAME[level] + " " + tag + message);
 };
 
 exports.ConsoleLogger = ConsoleLogger;
-exports.LEVEL = LEVEL;
+exports.LEVEL = baselogger.LEVEL;
+exports.LEVEL_COMPAT = baselogger.LEVEL_COMPAT;
```
wei-lee: I am wondering is there any benefit to keep this logger around, instead of just using either winston or bunyan logger instead.

@david-martin thoughts?

lib/sync-DataSetModel.js
```diff
@@ -574,8 +577,15 @@ var self = {
...
-  syncUtil.setLogger(syncUtil.SYNC_LOGGER, {logLevel: self.defaults.logLevel});
+var init = function(config) {
+
+  logger = config.logger;
+
+  if (!config || !config.logger) {
+    logger = new ConsoleLogger(baselogger.LEVEL_COMPAT[self.defaults.logLevel]);
+  }
+
+  syncUtil.setLogger(syncUtil.SYNC_LOGGER, logger);
```
wei-lee: this file will be removed, so we don't need to make any changes here.

lib/sync/index.js
wei-lee: +1. @david-martin I think you can incorporate this change into PR #101. This way we can access the event emitter in here and allow the sync framework to emit more events.

grdryn:
I like @wei-lee / @evanshortiss idea of using the `debug` module for the default. I don't have any opinions on whether that should be done as part of this issue, or a follow-up one, I guess that's for @wojta and @david-martin to decide. :)

lib/api.js
```diff
@@ -8,100 +8,93 @@ var sec = require('fh-security'),
...
-        act: 'hash',
-        params: opts
-      };
-      sec.security(p, callback);
-    },
-    web: require('./web')(cfg)
+  //
+  // Main FHapi constructor function..
+  //
+  function FHapi() {
```
evanshortiss: I don't think we'd ever make more than one instance so curious why we'd use proto here? Updates planned?

david-martin: @evanshortiss That change is from https://github.com/feedhenry/fh-mbaas-api/pull/101. It is using prototype so the api object being returned to the developer can extend EventEmitter.
However, this may be superceded by your PR #80 where `sync.events` is an EventEmitter

wojta:
I'm personally also for using some more general solution for the logger. @evanshortiss recommended https://github.com/evanshortiss/fh-bunyan. I was trying to do the change as careful as possible, so the logger API is the same and I don't break anything else outside Sync.

lib/api.js
evanshortiss: 👍 very nice

wojta: It's now removed, to incorporate changes in #101.

wojta:
Please recheck this.
Created follow up issue https://issues.jboss.org/browse/FH-3394 for implementing more common logger.

lib/sync/index.js
```diff
@@ -50,21 +53,35 @@ function connect(mongoDBConnectionUrl, redisUrl, metricsConf, cb) {
...
 }
 
 function setLogLevel(dataset_id, params, cb) {
   syncUtil.doLog(dataset_id, 'debug', 'setLogLevel');
-  if (params && params.logLevel) {
+  var logger = syncUtil.getLogger(dataset_id);
+  if (params && params.logLevel && logger) {
     syncUtil.doLog(dataset_id, 'debug', 'Setting logLevel to "' + params.logLevel + '"');
-    syncUtil.setLogger(dataset_id, params);
+    syncUtil.getLogger(dataset_id).setLogLevel(baselogger.LEVEL_COMPAT[params.logLevel]);
```
wojta: logger

david-martin:
Followup task to look into a universal logger implementation here https://issues.jboss.org/browse/FH-3394

lib/sync/api-sync.js
```diff
@@ -123,11 +122,11 @@ function processSyncAPI(datasetId, params, readDatasetClient, cb) {
  * @param {Function} cb the callback function
  */
 function sync(datasetId, params, cb) {
-  syncUtil.doLog(datasetId, 'debug', 'process sync request for dataset ' + datasetId);
+  debug('[%s] process sync request for the dataset');
```
david-martin: @wojta  missing `, datasetId` param?

lib/sync/api-syncRecords.js
```diff
@@ -170,21 +175,21 @@ function computeDelta(datasetId, clientRecords, serverRecords) {
  * @param {Function} cb
  */
 function syncRecords(datasetId, params, cb) {
-  syncUtil.doLog(datasetId, 'debug', 'process syncRecords request for dataset ' + datasetId);
+  debug('process syncRecords request for dataset %s', datasetId);
```
david-martin: should this follow the same pattern with `[%s]` for the dataset id?

lib/cache.js
```diff
@@ -1,12 +1,11 @@
 var redis = require("redis");
 var Memcached = require('memcached');
-
+var debug = require('debug')('fh-mbaas-api:debug');
```
david-martin: should this be `require('debug')('fh-mbaas-api:cache')`?

david-martin:
@wojta 1 minor eslint error in the CI build stopping progress
```
   lib/sync/storage/dataset-clients.js
      3 |var util = require('util');
                 ^ 'util' is defined but never used.
```

lib/cache.js
wojta: yes, I think so.

lib/sync/api-syncRecords.js
wojta: sure

lib/sync/api-sync.js
wojta: thx

lib/act.js
```diff
@@ -9,8 +9,9 @@ var call;
 var config;
 var widget;
 var cache;
+var debug = require('debug')('fh-mbaas-api:act');
```
evanshortiss: @wojta @david-martin  I would suggest that a wrapper for debug might be better than using the pure module? Logs with timestamps would be much more valuable.

```js
// debug-wrapper.js
var debug = require('debug')

module.exports = function getLogger(name) {
  var log = debug(name);

  return function doLog () {
    var args = Array.prototype.slice.call(arguments);

    args.unshift(new Date().toJSON()); // add a date stamp to each log

    log.apply(log, args);
  }
}
```

This print something like this:

```
testing 2017-03-24T15:16:51.370Z hello world
```

README.md
```diff
@@ -73,9 +73,9 @@ For example:
...
- where additional context is available that can't easily be passed to
- the calling code.  This idea comes from the
- [level suggestions for libraries in the Bunyan project](https://github.com/trentm/node-bunyan#level-suggestions).
- Users of the Sync API can then enable debug-level logging if they
- would like to see more output.
+Users of the Sync API can then enable debug-level logging if they would like to see more output. This is useful for debugging purposes.
+It's possible to pass environment variables to enable the logging according the rules specified for [debug](https://www.npmjs.com/package/debug) module:
+  
+```
+DEBUG="fh-mbaas-api:sync" DEBUG_COLORS=1 ./yourscript
```
evanshortiss: Might be worth specifying `fh-mbaas-api:*` will enable all logging?

lib/act.js
wojta: Timestamps are included automatically when you do DEBUG_COLORS=0, or it's running on non-TTY terminal. I don't know why they did that. 
I wouldn't include another wrapper if it's not necessary.

david-martin: I'd argue for simplicity over having a wrapper (just require and use it).

Maybe a good approach is forking the debug module & adding timestamps?
I think the current approach in the PR is a great improvment on what was there before, and could be built upon.
@wojta interested in your thoughts.

david-martin: I wrote above before your comment @wojta.
if debug can output timestamps, I'd be in favour of the approach already in the PR

README.md
wojta: ok, I'll include also note about timestamps in docs.

evanshortiss: Good call 👍  I can see people getting annoyed it they can't see any clear way to enable them.

lib/act.js
evanshortiss: I'd be in favor of not forking in that case and just documenting how to enable timestamps as @wojta says in the README comment.

david-martin: :+1:

PR #98, opened 2017-03-15
[FH 3228] split index js

paolobueno:
Moving a lot of setup and api implementation out of `lib/sync/index.js` into `lib/sync/sync-server.js`.
`index` is left with setting up metrics and the module's public API.

Currently pending a rebase on newer changes to the scaling branch and figuring out why some of the tests are still failing.

Putting up this WIP PR for preliminary review.

### DIFF
--- a/lib/sync/index.js
+++ b/lib/sync/index.js
@@ -14,18 +14,6 @@ function toJSON(dataset_id, returnData, cb) {
   return cb(null, {});
 }
 
-function listCollisions(dataset_id, params, cb) {
-  syncUtil.doLog(dataset_id, 'info', 'listCollisions');
-  return server.api.listCollisions(dataset_id, params.meta_data, cb);
-}
-
-// Defines a handler function for deleting a collision from the collisions list.
-// Should be called after the dataset is initialised.
-function removeCollision(dataset_id, params, cb) {
-  syncUtil.doLog(dataset_id, 'info', 'removeCollision');
-  return server.api.removeCollision(dataset_id, params.hash, params.meta_data, cb);
-}
-
 /**
  * Connect to mongodb & redis with the given connection urls.
  * This should be called before `start()` is called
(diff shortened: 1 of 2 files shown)


lib/sync/index.js
```diff
@@ -58,119 +12,16 @@ function toJSON(dataset_id, returnData, cb) {
...
-
 function listCollisions(dataset_id, params, cb) {
   syncUtil.doLog(dataset_id, 'info', 'listCollisions');
-  return dataHandlers.listCollisions(dataset_id, params.meta_data, cb);
+  return server.api.listCollisions(dataset_id, params.meta_data, cb);
 }
 
 // Defines a handler function for deleting a collision from the collisions list.
 // Should be called after the dataset is initialised.
 function removeCollision(dataset_id, params, cb) {
```
wei-lee: do we still need those 2 functions here? Perhaps we can remove them and just call the ones in `server.api`?

lib/sync/sync-server.js
```diff
@@ -0,0 +1,291 @@
...
+/**
+ * Connect to mongodb & redis with the given connection urls.
+ * This should be called before `start()` is called
+ *
+ * @param {string} mongoDBConnectionUrl A MongoDB connection uril in a format supported by the `mongodb` module
+ * @param {string} redisUrl A redis connection url in a format supported by the `redis` module
+ * @param {Object} metricsConf Metrics configuration
+ * @param {function} cb
+ */
+function connect(mongoDBConnectionUrl, redisUrl, metricsConf, cb) {
```
wei-lee: I am wondering should we leave this one in the `index.js` file. This function is mainly to setup the prerequisites before we starting the sync server. Feels like it's not part of the sync server.

lib/sync/index.js
paolobueno: I was keeping them here for two reasons:

- the extra logging line at the start (minor)
- the extracting of `params.meta_data` as it calls into the actual implementation

So if we move these two things into the function in `server.api` we would be able to keep the same API. I'm not very fond of having a function param just to extract a specific key from it, but I'll go ahead with this change.

PR #101, opened 2017-03-16
✨ FH-3366 Add sync:ready event to the fh api

david-martin:
FH-3366 Add sync_ready event to the fh api
This change required & incorporated a few things:
* adding a new `events` EventEmitter object to the fh API i.e. `fh.events`
* passing this into the sync server
* emitting a `sync:ready` event when sync is connected OK to mongo &
redis
* removing an uneccessary 'bootstrap' call for db string (& fix fhdb
test)

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -8,6 +8,7 @@ var sec = require('fh-security'),
 var url = require('url');
 var packageJSON = require('../package.json');
 var initScript = require('./init.js');
+var EventEmitter = require('events').EventEmitter;
 
 var mbaasClient = require('fh-mbaas-client');
 
@@ -29,6 +30,7 @@ function FHapi(cfg, logr) {
     },
     cache: require('./cache')(cfg),
     db: require('./db')(cfg),
+    events: new EventEmitter(),
     forms: require('./forms')(cfg),
     log: false,
     stringify: false,
@@ -56,6 +58,8 @@ function FHapi(cfg, logr) {
     },
     web: require('./web')(cfg)
   };
+
+  api.sync.setEventEmitter(api.events);
 
   var redisUrl = 'redis://' + api.redisHost + ':' + api.redisPort;
 
@@ -87,16 +91,6 @@ function FHapi(cfg, logr) {
     // Call stopAll to ensure Sync exits clenaly.
     api.sync.stopAll(cb);
   };
-
-  api.bootstrap = function(cb) {
-    api.db({'act': 'connectionString'}, cb);
-  };
-
-  api.bootstrap(function(err) {
-    if (err) {
-      logr.error(err);
-    }
-  });
 
   api.setLogLevel = function(level) {
     logr.setLogLevel(level);
@@ -220,5 +214,5 @@ module.exports = (function () {
 
   var logger = new consolelogger.ConsoleLogger(logLevel);
   initScript(logger);
-  return new FHapi(cfg, logger);
+  return FHapi(cfg, logger);
 })();
--- a/lib/sync/index.js
+++ b/lib/sync/index.js
@@ -6,7 +6,7 @@ var util = require('util');
 var MongoClient = require('mongodb').MongoClient;
 var redis = require('redis');
 var async = require('async');
-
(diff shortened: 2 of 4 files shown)


david-martin:
@wei-lee I would appreciate your thoughts on this change, and how it is implemented.
More info on the reasons for this change are in https://issues.jboss.org/browse/FH-3366

lib/api.js
```diff
@@ -8,100 +8,93 @@ var sec = require('fh-security'),
...
+  FHapi.prototype.service = require('./act')(cfg);
+  FHapi.prototype.auth = require('./auth')(cfg);
+  FHapi.prototype.host = require('./host');
+  FHapi.prototype.permission_map = require('fh-db').permission_map;
+  FHapi.prototype.hash = function hash(opts, callback) {
+    var p = {
+      act: 'hash',
+      params: opts
+    };
+    sec.security(p, callback);
```
wei-lee: `sec` is not defined. Should be `this.sec`

lib/api.js
```diff
@@ -8,100 +8,93 @@ var sec = require('fh-security'),
...
+  },
+  FHapi.prototype.mbaasExpress = function (opts) {
+    opts = opts || {};
+    opts.api = this;
+    return require('fh-mbaas-express')(opts);
+  };
+  FHapi.prototype.shutdown = function (cb) {
+    // Sync service has a setInterval loop running which will prevent fh-mbaas-api from exiting cleanly.
+    // Call stopAll to ensure Sync exits clenaly.
+    this.sync.stopAll(cb);
```
wei-lee: where is `this.sync` defined?

lib/api.js
wei-lee: never mind. I saw `sec` is the module.

lib/api.js
david-martin: I believe it should be the sync function on the prototype defined above
```
FHapi.prototype.sync = require('./sync').api;
```

lib/api.js
```diff
@@ -8,100 +8,93 @@ var sec = require('fh-security'),
...
       return;
     }
+
+    var redisUrl = 'redis://' + api.redisHost + ':' + api.redisPort;
     api.sync.connect(connectionString, redisUrl, null, function(err) {
       if (err) {
         console.error('Error starting the sync server (', err, ')');
+        return;
       }
+      api.emit('sync_ready');
```
wei-lee: given that this event is really about the sync framework itself, I am wondering would it better if the sync api itself emit the event, rather than do it from here?

david-martin: @wei-lee Good point. After a successful `connect`, sync could emit it.
This would follow the pattern proposed by @evanshortiss in PR #80 where an event emitter is passed to sync, and that event emitter is initialised up at the fh-mbaas-api level

david-martin:
@evanshortiss I've reset this branch and taken an approach similar to your PR #80.
Woudl appreciative you're feedback.

@wei-lee I've used `sync:ready` as the event name to use the same namespacing pattern as raincatcher.

The API change for developers is a new `events` EventEmitter on the exported object e.g.
```js
fh.events.on('sync:ready', function() {})
```
This object is passed into the sync server, which uses it to emit the event.
The same eventEmitter could be passed into other parts of the API as needed.

lib/sync/index.js
```diff
@@ -65,6 +66,16 @@ function setLogLevel(dataset_id, params, cb) {
   }
 }
 
+function getEventEmitter(emitter) {
```
wei-lee: the `emitter` param is not required

david-martin: Good spot. Removed

evanshortiss:
@wei-lee @david-martin any changes you'd like to my PR to work with this? I need to make the change to the collision handler to take a callback still, but once that's done my PR is likely finished unless more events are necessary, such as sync-ready?

david-martin:
@evanshortiss If you're referring to PR #80 (I closed it recently), you may want to rebase/redo it on top of FH-3123-sync-horizontal-scaling so your suggested changes are applied on top of the new sync implementation

PR #102, opened 2017-03-21
add an isLocal method

evanshortiss:
Open to opinions on whether this belongs here or not. Thought it would be useful vs. having people check for different env vars in their code.

Sample use case:

```js
var mbaasApi = require('fh-mbaas-api');
var createLocalData = require('fixtures/create');

if (mbaasApi.isLocal()) {
  createLocalData.then(startApp);
} else {
  startApp()
}
```

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -69,6 +69,9 @@ function FHapi(cfg, logr) {
     api.sync.stopAll(cb);
   };
 
+
+  api.isLocal = require('./fhutils').isLocal;
+
   api.bootstrap = function(cb) {
     api.db({'act': 'connectionString'}, cb);
   };
--- a/lib/fhutils.js
+++ b/lib/fhutils.js
@@ -1,4 +1,5 @@
 var assert = require('assert');
+var process = require('process');
 
 module.exports = function (cfg) {
   assert.ok(cfg, 'cfg is undefined');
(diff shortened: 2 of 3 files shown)


david-martin:
@evanshortiss I can understand your use case for `isLocal`, but I'm wondering if it's opening the possibility of developers writing potentially dangerous code.
If for some reason `isLocal` started to behave incorrectly, it could mean an app running in a production scenario would behave unexpectedly.
To take your example, if `isLocal` started to think it was running locally due to missing env vars, the production database would have whatever `createLocalData` does applied to it.

I think config or environment settings are the way to solve this, but I wonder if it should be controlled outside the app.
How about an extra env var being set by grunt when you do `grunt serve` locally e.g. `CREATE_LOCAL_DATA=1`. It might be better to namespace as well just in case e.g. `MYAPP_CREATE_LOCAL_DATA=1`

evanshortiss:
Using custom vars is a good solution too. My thinking was that we already use FH_USE_LOCAL_DB and FH_SERVICE_MAP (I forgot about this) in our templates and modules already so having rather than writing a `utils.isLocal` for each app it'd be nice to have it out of the box. 

I can understand that taking this risk into the SDK is not desirable though! Closing

PR #106, opened 2017-03-23
add new function to return status info about the sync framework

wei-lee:
This PR will add a new function that will return the stats info about the sync framework.

The stats info will be collected automatically by default and saved in Redis. The new function will then fetch those metrics from redis and produce the summary data for the users.


wei-lee:
Retest this please

wei-lee:
@aidenkeating  we want to allow users to pass in the mongodb connection options when `connect` is called, especially to specify the 'poolSize'.  see https://github.com/wei-lee/fh-mbaas-api-1/blob/e9accd5b0c8c4e7bfc4e23bed845ca8306de7c1a/lib/api.js#L76. It will help improve performance.

PR #108, opened 2017-03-24
FH-3242 remove old sync code

wojta:
Old code was removed.


wojta:
@david-martin @wei-lee someone pls check this

PR #109, opened 2017-03-27
🐛 FH-3413 Revert collision handler function interface to maintain backwards compatibility

david-martin:
See https://issues.jboss.org/browse/FH-3413 for more info

### DIFF
--- a/test/sync/test_dataHandlers.js
+++ b/test/sync/test_dataHandlers.js
@@ -90,11 +90,10 @@ module.exports = {
 
   'test set custom collisionHandler': function(done) {
     var dataHandlers = dataHandlersModule({defaultHandlers: defaultHandlers});
-    dataHandlers.collisionHandler(id, function (datasetId, metadata, collisionFields, cb) {
+    dataHandlers.collisionHandler(id, function(dataset_id, hash, timestamp, uid, pre, post, meta_data, cb) {
       cb(null);
     });
-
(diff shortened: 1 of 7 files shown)


david-martin:
@aidenkeating @wei-lee Could you take a look?

test/sync/test_dataHandlers.js
```diff
@@ -142,7 +142,7 @@ module.exports = {
 
   'test set default collisionHandler': function(done) {
     var dataHandlers = dataHandlersModule({defaultHandlers: defaultHandlers});
-    dataHandlers.handleCollision(id, metaData, [],  function(err) {
+    dataHandlers.handleCollision(id, metaData, [], function(err) {
```
wei-lee: should we change the signature here?

david-martin: @wei-lee I didn't think so as it looked to be calling a layer between this and the actual handler.
https://github.com/feedhenry/fh-mbaas-api/blob/FH-3123-sync-horizontal-scaling/lib/sync/dataHandlers.js#L236-L239
However, now I'm not sure.
I think it might be best to change the signature everywhere?

wei-lee: I should be changed everywhere. That function should invoke the default `handleCollision` function.

david-martin: @wei-lee Thanks. Updated remaining function calls

PR #110, opened 2017-03-31
FH-3417 Ensure clients update when there is a collision

aidenkeating:
JIRA: https://issues.jboss.org/browse/FH-3417

I tried explaining it all in the commit message, but it probably makes things more complicated.

### DIFF
--- a/lib/sync/dataHandlers.js
+++ b/lib/sync/dataHandlers.js
@@ -13,7 +13,8 @@ module.exports = function(options) {
     DELETE: 'deleteHandler',
     COLLISION: 'collisionHandler',
     LIST_COLLISIONS: 'listCollisionsHandler',
-    REMOVE_COLLISION: 'removeCollisionHandler'
+    REMOVE_COLLISION: 'removeCollisionHandler',
+    COUNT_COLLISIONS: 'countCollisionsHandler'
   }
   var handlers;
   var globalHandlers;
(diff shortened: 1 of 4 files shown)


david-martin:
The test failing looks like a missing mock/spy.
I think we could really benefit from an automated test for this scenario in https://github.com/feedhenry/sync-acceptance-testing as its a complex scenario to setup manually

lib/sync/dataHandlers.js
```diff
@@ -259,6 +261,16 @@ module.exports = function(options) {
       handler(datasetId, dataHash, metaData, callback);
     },
 
+    /**
+     * @param datasetId
+     * @param metaData
+     * @param callback
+     */
+    countCollisions: function(datasetId, metaData, callback) {
```
wei-lee: I think there are some problem here:

1. it adds a new function that the clients needs to override
2. it could return duplicated value - if some of the collisions are resolved.

I think a simpler approach is to just keep an counter on the dataset client, and then every time when there is a collision, we can increase that counter.

aidenkeating: @wei-lee I'm probably missing something but I can't find a place where this should/could be done? We find if there is a collision in the pending worker, but we don't seem to have a way to access the dataset client from there in order to update it?

wei-lee: @aidenkeating a collision is generated from the pending processor, for example: https://github.com/feedhenry/fh-mbaas-api/blob/FH-3123-sync-horizontal-scaling/lib/sync/pending-processor.js#L39.

We can just increment the collision counter at this point, if we keep it on the dataset client document. This way we don't need to count the collisions at all.

aidenkeating:
@wei-lee Thanks for the feedback, I've opened the changes as a new PR here: https://github.com/feedhenry/fh-mbaas-api/pull/112

PR #111, opened 2017-03-31
Fh 3433 worker backoff

wei-lee:
This PR add the following improvements:

1. use the `collectStatsInterval` configuration of the sync server to determine when to report the queue size from workers.
2. Allow workers to backoff if there are no jobs to process. This will reduce the resource usages by workers when there are no jobs to process.


lib/sync/sync-server.js
```diff
@@ -45,10 +45,16 @@ var syncWorker;
 var DEFAULT_SYNC_CONF = {
   /** @type {Number} how often pending workers should check for the next job, in ms. Default: 500 */
   pendingWorkerInterval: 500,
+  /** @type {Object} the backoff strategy for the pending worker to use. Default is exponetial, and the max delay is 60s */
```
david-martin: using 'max delay' in this comment makes it sound like you can't set the max over 60s.

Maybe something like this could read better?
'The default strategy is "exp" with a max delay of 60 seconds'

lib/sync/sync-server.js
```diff
@@ -45,10 +45,16 @@ var syncWorker;
...
+  pendingWorkerBackoff: {strategy: 'exp', max: 60*1000},
   /** @type {Number} how often ack workers should check for the next job, in ms. Default: 500 */
   ackWorkerInterval: 500,
+  /** @type {Object} the backoff strategy for the ack worker to use. Default is exponetial, and the max delay is 60s */
```
david-martin: We should inlcude some info on what the different strategies are, and what exactly 'max delay' means

lib/sync/worker.js
```diff
@@ -1,6 +1,27 @@
...
+var backoff = require('backoff');
+
+function initBackoff(backOffOpts) {
+  var bo;
+  if (backOffOpts.strategy === 'exp') {
+    bo = new backoff.ExponentialStrategy({initialDelay: backOffOpts.min, maxDelay: backOffOpts.max});
+  } else if (backOffOpts.strategy === 'fib') {
+    bo = new backoff.FibonacciStrategy({initialDelay: backOffOpts.min, maxDelay: backOffOpts.max});
+  } else {
+    //if no valid value is given, just return the default minimum delay. Can be used to turn off backoff.
```
david-martin: Should this throw an exception if an invalid strategy is configured?
It might be safer that a misspelled config. For example if someone sets the strategy to "FIB", it won't use fibonnaci. Instead it will use the 'minimum' strategy. This might be a cause of confusion

lib/sync/sync-server.js
david-martin: Also, worth mentioning 'min' too

test/sync/test_worker.js
```diff
@@ -34,9 +34,10 @@ module.exports = {
     q.get.onThirdCall().yields(null, null);
     q.get.onCall(3).yields(null, task2);
     q.ack.yields(new Error('test ack failed'));
-    var worker = new Worker(q, processor, metrics, {interval: 10});
+    var worker = new Worker(q, processor, metrics, {interval: 10, backoff: {strategy: 'none'}});
```
david-martin: Is 'none' a special value, or just depending on it not matching any known strategy?

david-martin:
@wei-lee Reviewed with some comments.

lib/sync/sync-server.js
wei-lee: good point. Will update this.

lib/sync/worker.js
wei-lee: I think it should ignore the case. I don't think thrown an exception here is right as it doesn't really affect the running of the apps even there is no backoff. I think we can print out a warning message here to make it obvious that the default strategy is being used.

Given that there is only 3 charactors, I think it's hard to mis-spell.

test/sync/test_worker.js
wei-lee: no, it's not special. It's not matching any of the strategy values.

PR #112, opened 2017-03-31
FH-3417 ensure update on collision

aidenkeating:


lib/sync/DatasetClient.js
```diff
@@ -128,6 +130,13 @@ DatasetClient.prototype.setGlobalHash = function(globalHash) {
 };
 
 /**
+ * Get the current count of collisions that have happened on the dataset client
+ */
+DatasetClient.prototype.getCollisionCount = function() {
```
wei-lee: this doesn't seem like to be used here.

lib/sync/pending-processor.js
```diff
@@ -36,7 +36,13 @@ function handleCollision(datasetId, metaData, pendingChange, dataHash, callback)
     post: pendingChange.post,
     timestamp: pendingChange.timestamp
   };
-  dataHandlers.handleCollision(datasetId, collisionFields.hash, collisionFields.timestamp, collisionFields.uid, collisionFields.pre, collisionFields.post, metaData, callback);
+  var newCollisionCount = pendingChange.datasetClient.collisionCount + 1;
+  syncStorage.updateDatasetClient(pendingChange.datasetClient.id, {collisionCount: newCollisionCount}, function(err) {
+    if (err) {
+      debug('[%s] Updating dataset client collision count (%s) failed : err = %j', datasetId, pendingChange.datasetClient.id, err);
+    }
+    dataHandlers.handleCollision(datasetId, collisionFields.hash, collisionFields.timestamp, collisionFields.uid, collisionFields.pre, collisionFields.post, metaData, callback);
```
wei-lee: I think those 2 calls can happen in parallel?

lib/sync/sync-processor.js
```diff
@@ -95,15 +95,18 @@ function syncWithBackend(payload, tries, callback) {
...
+    function listData(datasetClient, cb) {
+      listDataWithTimeout({datasetClientId: datasetClientId, datasetId: datasetId, queryParams: queryParams, metaData: metaData}, expectedTimeout, function(err, res) {
+        return cb(err, datasetClient, res);
+      });
     },
-    function saveRecords(recordsWithHash, cb) {
-      var globalHash = hashProvider.globalHash(datasetId, _.pluck(recordsWithHash, 'hash'));
+    function saveRecords(datasetClient, recordsWithHash, cb) {
+      var toHash = _.pluck(recordsWithHash, 'hash').concat([datasetClient.collisionCount]);
+      var globalHash = hashProvider.globalHash(datasetId, toHash);
```
wei-lee: I am wondering should we just suffix the collision count value to the global hash, rather than add it to the records hashes and then hash it, so the new global hash will be like:

`23423498hsdkkf_1`

The reason is that it will allow us to see which is causing the global hash to change - is it because of records changes, or because of collision changes. This way will make it easy to find out.

WDYT?

wei-lee:
a few comments, overall LGTM

lib/sync/sync-processor.js
aidenkeating: @wei-lee Yes, I think that makes sense.

aidenkeating:
@wei-lee I think that's everything?

aidenkeating:
Ooh, but we probably shouldn't merge before I add in these acceptance tests

david-martin:
retest this please

aidenkeating:
# Top Banana!
Top Banana

@david-martin Good to merge?

PR #113, opened 2017-04-04
Fh 3435 add new app

wei-lee:
This PR:

1. introduces a new env var `SYNC_MONGODB_URL`. If this value is set, the sync server will use it as the mongodb connection string. This will allow other apps to connect to the mongodb of an existing app.
2. Introduce concurrencies for workers. Multiple workers of the same type can be created inside an app.
3. Disable the default backoff behaviour of the sync worker. This is to ensure the sync frequency will work. Otherwise the backoff will make the sync frequency un-predicated.

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -68,6 +68,7 @@ function FHapi(cfg) {
       return;
     }
     var poolSize = parseInt(process.env.SYNC_MONGODB_POOLSIZE) || 50;
+    console.log('sync.connect with mongodb = %s and redis = %s', connectionString, redisUrl);
     api.sync.connect(connectionString, {poolSize: poolSize}, redisUrl, function(err) {
       if (err) {
         console.error('Error starting the sync server (', err, ')');
--- a/lib/api.js
+++ b/lib/api.js
@@ -12,6 +12,16 @@ var mbaasClient = require('fh-mbaas-client');
 //IMPORTANT: This will force node to ignore cert errors for https requests
 process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0";
 
+function getDBUrl(api, callback) {
+  if (process.env.SYNC_MONGODB_URL) {
+    return callback(null, process.env.SYNC_MONGODB_URL);
+  } else {
+    api.db({
+      "act" : "connectionString"
+    }, callback);
+  }
+}
+
 //
 // Main FHapi constructor function..
 //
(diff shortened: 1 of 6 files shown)


wei-lee:

lib/api.js
```diff
@@ -68,6 +68,7 @@ function FHapi(cfg) {
       return;
     }
     var poolSize = parseInt(process.env.SYNC_MONGODB_POOLSIZE) || 50;
+    console.log('sync.connect with mongodb = %s and redis = %s', connectionString, redisUrl);
```
david-martin: @wei-lee should probably use 'debug' here

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "fh-mbaas-api",
+  "name": "fh-mbaas-api-metrics",
```
david-martin: did you mean to make this change?

david-martin:
@wei-lee These changes look reasonable.
A couple comments above.
Also, is there a way we can test these changes (unit and/or integration)?

wei-lee:
@david-martin the comments you added are for an old commit I think. They are not in the PR anymore.

The existing integration tests already cover the changes around start/stop.

wei-lee:
@david-martin ok to merge?

lib/sync/sync-server.js
```diff
@@ -356,6 +356,8 @@ module.exports = {
     listCollisions: listCollisions,
     removeCollision: removeCollision,
     callHandler: callHandler,
-    getStats: getStats
+    getStats: getStats,
+    setRecordHashFn: hashProviderModule.setRecordHashFn,
```
david-martin: are these 2 functions new to the API?

david-martin: Also, this doesn't look related to the goal of the PR. 
We can include it if you want, but would be good to get more context

david-martin:
@wei-lee  Merging now after conversation on IRC

```
<irldavem> weili_, re: https://github.com/feedhenry/fh-mbaas-api/pull/113, i don't mind if you include the hashing funcitons
<irldavem> but are the names of them correct?
<irldavem> i thought they were in the existing sync server too?
<irldavem> from Evans or Philips pr
<weili_> irldavem yes, but the existing one the name is `globalHandleHash`
<weili_> irldavem which is a bit confusing
<weili_> irldavem given that they are not yet documented
<weili_> irldavem I think we can take the opportunity to rename it
<irldavem> weili_, so document that they are added in 7.0.0?
<weili_> irldavem yes
```

PR #119, opened 2017-04-10
 🎨 start the datasetClient cleaner job to remove the inactive dataset…

wei-lee:
…Clients

This PR will make sure the datasetClient cleaner will be started and stopped when the sync server is started or stopped.

A datasetClientCleaner will delete all the inactive datasetClients that are older than then given time threshold.


wei-lee:
ping @david-martin  @aidenkeating for review.

PR #121, opened 2017-04-11
 🎨 add a new configuration option to override the metrics client repo…

wei-lee:
…rt concurrency

### DIFF
--- a/lib/sync/sync-metrics.js
+++ b/lib/sync/sync-metrics.js
@@ -182,7 +182,8 @@ module.exports = {
         type: 'redis',
         redisClient: redisClientImpl,
         namespace: statsNamespace,
-        recordsToKeep: syncConfig.statsRecordsToKeep
+        recordsToKeep: syncConfig.statsRecordsToKeep,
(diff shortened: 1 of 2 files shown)


lib/sync/sync-metrics.js
```diff
@@ -191,7 +192,8 @@ module.exports = {
       metricsConfig.backends.push({
         type: 'influxdb',
         host: syncConfig.metricsInfluxdbHost,
-        port: syncConfig.metricsInfluxdbPort
+        port: syncConfig.metricsInfluxdbPort,
+        sendQConcurrency: syncConfig.metricsReportConcurrency
```
aidenkeating: Is this option not `sendQueueConcurrency`?  I think `sendQConcurrency` is only used internally.

https://github.com/feedhenry/fh-component-metrics/blob/b8565be0205aed2f8eafe3cfd8be615db5c3d679/lib/clients/base.js#L9

wei-lee: yes, good catch. Will fix it

PR #122, opened 2017-04-12
Change default dataset sync frequency to match clients

aidenkeating:


david-martin:
@aidenkeating This makes sense to me given it was the default value in the older sync implementation.
Also, as you say in your commit message, the default sync on client templates is 10, which makes sense to match as a sensible default.

PR #129, opened 2017-04-26
Fix timeout errors

wei-lee:
This PR:

1. Do not wait for the pending changes to be processed before returning response for sync API call. Instead return data ASAP, and process pending changes in `process.nextTick`.
2. Remove references to `datasetClientRecords`, it's not used and could use a lot of memory.

@david-martin can you please review the change?

### DIFF
--- a/lib/sync-srv.js
+++ b/lib/sync-srv.js
@@ -165,16 +165,22 @@ function doClientSync(dataset_id, params, callback) {
         if (params.pending && params.pending.length > 0) {
           syncUtil.doLog(dataset_id, 'info', 'Found ' + params.pending.length + ' pending records. processing', params);
 
-          // Process Pending Params then re-sync data
-          processPending(dataset_id, dataset, params, function () {
-            syncUtil.doLog(dataset_id, 'verbose', 'back from processPending', params);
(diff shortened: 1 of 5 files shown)


david-martin:
Taking review of this

lib/sync-srv.js
```diff
@@ -183,7 +189,7 @@ function doClientSync(dataset_id, params, callback) {
             var res;
             if (datasetClient.dataHash === params.dataset_hash) {
               syncUtil.doLog(dataset_id, 'verbose', 'doClientSync - No pending - Hashes match. Just return hash', params);
-              res = {"hash": datasetClient.hash};
+              res = {"hash": datasetClient.dataHash};
```
david-martin: @wei-lee is this an unrelated bugfix?

aidenkeating:
@wei-lee I think there must be some issue with the collision tests with `6.x` sync as I can produce the collision using the same steps locally.

lib/sync-srv.js
wei-lee: @david-martin yes, it should be `dataHash`

david-martin:
closing as out of date

PR #130, opened 2017-04-28
Fix memory issue

wei-lee:
## Why

During  load testing, we discovered that the app is using too much memory and CPU. After investigation, it turns out there the load test generate about 1700 datasetClients in the db, and some of those datasetClients have a lot of records associated with it.

In the syncScheduler, to check if a datasetClient should be synced, we do a `collection.list({})` operation. This operation lists ALL the data records and ALL the data fields in to memory and then process them. Since the collection data is relatively large, this has caused memory issue and CPU issue (to parse all the data).

To make things worse, there is a datasetClientCleaner that will run at the start of the app. It tries to identify the datasetClients that haven't been active for too long and remove them. It is listing all the datasetClients as well.
 
## What

To fix the issue, we add list filters to only get the datasetClients we need, and don't list the fields we don't need:

1. when list datasetClients in the syncScheduler, we only list `active` datasetClients. When those datasetClients are being checked, some of them will have the `active` field set to false if they are not active. The `active` field will be set to `true` again once they are accessed again from clients.
2. when cleanup datasetClients, we only need to list the ones that are not active.

Also this PR fixed an issue when a datasetClient could be removed before it's being marked as "inactive".

ping @david-martin  @aidenkeating for review

### DIFF
--- a/lib/sync/storage/dataset-clients.js
+++ b/lib/sync/storage/dataset-clients.js
@@ -31,14 +31,15 @@ var cacheClient;
 
 /**
  * List all the dataset clients docs.
+ * @param {Object} filter
  * @param {Function} cb
  */
-function doListDatasetClients(cb) {
+function doListDatasetClients(filter, cb) {
   debug('doListDatasetClients');
   var col = mongoClient.collection(DATASETCLIENTS_COLLECTION);
-  col.find({}).toArray(function(err, datasetClients) {
+  col.find(filter).project({recordUids: 0}).toArray(function(err, datasetClients) {
     if (err) {
-      debug('Failed to list datasetClients due to error %j', err);
+      debug('Failed to list doListDatasetClients due to error %j', err);
     }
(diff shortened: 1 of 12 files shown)


lib/sync/storage/dataset-clients.js
```diff
@@ -31,14 +31,15 @@ var cacheClient;
...
  */
-function doListDatasetClients(cb) {
+function doListDatasetClients(filter, cb) {
   debug('doListDatasetClients');
   var col = mongoClient.collection(DATASETCLIENTS_COLLECTION);
-  col.find({}).toArray(function(err, datasetClients) {
+  col.find(filter).project({recordUids: 0}).toArray(function(err, datasetClients) {
     if (err) {
-      debug('Failed to list datasetClients due to error %j', err);
+      debug('Failed to list doListDatasetClients due to error %j', err);
```
david-martin: minor copy/paste in debug statement here

PR #131, opened 2017-04-28
🐛 RHMAP-15177 Fix an issue for when FH_USE_LOCAL_DB is set

david-martin:
If FH_USE_LOCAL_DB is set, also set the FH_MONGODB_CONN_URL env var
to a localhost string so sync can start.

This also addresses RHMAP-15175

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -63,6 +63,10 @@ function getMongoConnectionUrl(cb) {
     return mongoConnectionStringOS2(cb);
   } else if ("openshift3" === process.env.FH_MBAAS_TYPE) {
     return mongoConnectionStringOS3(cb);
+  } else if (process.env.FH_USE_LOCAL_DB) {
+    // Backwards compatible, hard coded localhost for local development
+    process.env.FH_MONGODB_CONN_URL = 'mongodb://localhost:27017/FH_LOCAL';
+    return cb(undefined, process.env.FH_MONGODB_CONN_URL);
   } else {
     debug('no way to determine mongo connection string');
     return cb();
@@ -117,7 +121,7 @@ var db = function(params, callback) {
     }
 
     var fhdb = require('fh-db');
-    if (process.env.FH_USE_LOCAL_DB || process.env.FH_MONGODB_CONN_URL) {
+    if (process.env.FH_MONGODB_CONN_URL) {
       return fhdb.local_db(params, callback);
     }
     else {
--- a/lib/sync/sync-server.js
+++ b/lib/sync/sync-server.js
@@ -153,15 +153,10 @@ function stopAllWorkers(workers, cb) {
 function start(cb) {
   if (arguments.length < 1) throw new Error('start requires 1 argument');
 
-  if (syncStarted) {
-    return cb();
-  } else {
-    //prevent `start` getting called multiple times
-    syncStarted = true;
-  }
+  syncStarted = true;
 
   if (mongoDbClient === null || redisClient === null) {
-    return cb('MongoDB Client & Redis Client are not connected. Ensure connect() is called before calling start');
+    return cb(new Error('MongoDB Client & Redis Client are not connected. Ensure connect() is called before calling start'));
   }
 
   metricsClient = metricsModule.init(syncConfig, redisClient);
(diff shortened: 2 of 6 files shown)


lib/db.js
```diff
@@ -63,6 +63,10 @@ function getMongoConnectionUrl(cb) {
     return mongoConnectionStringOS2(cb);
   } else if ("openshift3" === process.env.FH_MBAAS_TYPE) {
     return mongoConnectionStringOS3(cb);
+  } else if (process.env.FH_USE_LOCAL_DB) {
+    // Backwards compatible, hard coded localhost for local development
+    process.env.FH_MONGODB_CONN_URL = 'mongodb://localhost:27017';
```
wei-lee: is it possible the mongodb url maybe different locally? Maybe check if there is another mongodb url env var first and then default to the local one? WDYT?

aidenkeating: @wei-lee I might be misunderstanding but isn't that what's done in the logic above this? I would have thought if the mongodb url was different locally the user could set the `FH_MONGODB_CONN_URL` env var? Then this logic will never get used

@david-martin Should we be using the `admin` database by default? Or would it be possible to use a named database? Like how we would usually use `FH_LOCAL`

david-martin: @aidenkeating yes, i'll change this to FH_LOCAL to avoid it using the admin database

david-martin:
depends on fh-db change here https://github.com/feedhenry/fh-db/pull/27

lib/sync/sync-server.js
```diff
@@ -153,15 +153,10 @@ function stopAllWorkers(workers, cb) {
 function start(cb) {
   if (arguments.length < 1) throw new Error('start requires 1 argument');
 
-  if (syncStarted) {
-    return cb();
-  } else {
-    //prevent `start` getting called multiple times
-    syncStarted = true;
-  }
+  syncStarted = true;
```
wei-lee: if we are using `async. memoize`, if `syncStarted` still required?

lib/sync/sync-server.js
```diff
@@ -361,7 +358,8 @@ module.exports = {
   setClients: setClients,
   api: {
     init: init,
-    start: start,
+    // Memoize the start function so it will only get called at most once
+    start: async.memoize(start),
```
wei-lee: are we sure the function will be only called once, even if the second call happens before the first call is completed? It is not that obvious from the description of documentation.

wei-lee:
I have a few question about using `async. memoize`, not sure about the benefit of using it

lib/sync/sync-server.js
david-martin: @wei-lee Positive.
See this example

```js
var async = require('async');

var slow_fn = function(callback) {
    console.log(new Date().toString(), 'calling slow_fn');
    setTimeout(function() {
	    console.log(new Date().toString(), 'slow_fn callback');
	    callback(null, 'done');
    }, 2000);
};
var fn = async.memoize(slow_fn);
var cb = function(err, val) {
  console.log(new Date().toString(), val);
};

fn(cb);
fn(cb);
fn(cb);
fn(cb);
```

```
node test.js 
Tue May 02 2017 16:24:38 GMT+0100 (IST) calling slow_fn
Tue May 02 2017 16:24:40 GMT+0100 (IST) slow_fn callback
Tue May 02 2017 16:24:40 GMT+0100 (IST) done
Tue May 02 2017 16:24:40 GMT+0100 (IST) done
Tue May 02 2017 16:24:40 GMT+0100 (IST) done
Tue May 02 2017 16:24:40 GMT+0100 (IST) done
```

lib/sync/sync-server.js
david-martin: syncStarted is not required. I'll remove

david-martin:
@wei-lee The benefit is to prevent this issue https://issues.jboss.org/browse/FH-3555.
Do you think the bug is invalid?

lib/sync/sync-server.js
wei-lee: excellent

lib/sync/sync-server.js
david-martin: @wei-lee actually I'll keep this flag as its being checked when 'stop' is called and behaves differently if sync was already started.

PR #132, opened 2017-05-02
Add `fh-mbaas-api:sync:error` debug level

aidenkeating:
Currently we have one log level for sync. This causes a lot of logs to be
generated when the app is running. This adds an extra `error` level to the
debug logs so only error messages will be output. This should result in logs
in the studio not rotating so often, making issues easier to debug when you
are only concerned with errors.

This reduces the overall amount of logs considerably.
This also adds the same `:error` level for all other debug logs (e.g. stats).

To log all errors in `fh-mbaas-api` you can use `DEBUG=fh-mbaas-api:*:error`

### DIFF
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -1,6 +1,6 @@
 var redis = require("redis");
 var Memcached = require('memcached');
-var debug = require('debug')('fh-mbaas-api:cache');
+var debugError = require('debug')('fh-mbaas-api:cache:error');
 var redisPort = 6379;
 var redisHost = '127.0.0.1';
 var redisPassword;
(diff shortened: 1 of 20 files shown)


aidenkeating:
@wei-lee @david-martin Would you mind taking a look?

david-martin:
@aidenkeating I'm ok with allowing filtering of logs like this.
However, I wonder if any of the 'errors' being supressed by default are actually errors that the developer needs to know about?
If there are errors, we should probably get the errors back to the user.
For any that aren't really errors and are more informative (safe for developer to ignore), these might be better worded as warnings.
wdyt?

wei-lee:
+1

aidenkeating:
@david-martin Looking back over the changes, all of the errors that occur (in sync at least) are returned to the user through a callback. Based on the comments here https://github.com/feedhenry/fh-mbaas-api/pull/133#issuecomment-298879914 I am wondering whether these `debugError` statements should be changed to `console.error`?

aidenkeating:
@david-martin @wei-lee This has been updated to use `console.error` statements, would you mind taking a look?

wei-lee:
I think you may want to check the string formatting. Some of the placeholders are unique to the `debug` module, but not supported by console.error. Like `%j`. Otherwise looks ok.

aidenkeating:
@wei-lee After reading further it looks like `%j` is supported in Node's `console.error`. It uses `util.format` underneath. Is this good to merge?

wei-lee:
@aidenkeating good to know

PR #133, opened 2017-05-03
Rhmap 15214 - improve performance when delete datasetClients

wei-lee:
## Why

At the moment, when datasetClients are being deleted, we need to read all the datasetClient records that need to be deleted into memory, loop through them, and loop through each item in the `recordUids` array, find the corresponding record and update/remove each of them.

However, each datasetClient could have many `recordUids`. If all of the datasetClients are loaded into memory and processed at the same time, it could cause memory issue for the app. We can also process each datasetClient one by one, but the whole process will be too slow.

After a bit of investigation, I realise we actually don't need the `recordUids` fields on the datasetClients collection at all. This is because each record is keep tracking what datasetClients are referencing it via the `refs` field. In Mongodb, we can add an index on an array field and we will be able to find out what records belong to a datasetClient quite easily via a single Mongodb query. We can also perform batch updates and removes quite easily on the records collection.

## What

* Remove the `recordsUid` field from datasetClient collection
* Change the implementation for `removeDatasetClient`. Now it becomes:
** find and update all the records belong to the datasetClient to remove the reference in a single Mongodb query
** remove the datasetClient record
** remove all the unused records in a single query

### DIFF
--- a/lib/sync/storage/dataset-clients.js
+++ b/lib/sync/storage/dataset-clients.js
@@ -37,7 +37,7 @@ var cacheClient;
 function doListDatasetClients(filter, cb) {
   debug('doListDatasetClients');
   var col = mongoClient.collection(DATASETCLIENTS_COLLECTION);
-  col.find(filter).project({recordUids: 0}).toArray(function(err, datasetClients) {
+  col.find(filter).toArray(function(err, datasetClients) {
     if (err) {
       debug('Failed to list datasetClients due to error %j', err);
     }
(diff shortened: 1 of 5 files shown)


lib/sync/storage/dataset-clients.js
```diff
@@ -52,12 +52,31 @@ function doListDatasetClients(filter, cb) {
...
   var removeIds = _.pluck(datasetClientsToRemove, 'id');
-  debug('doRemoveDatasetClients removeIds = %j', removeIds);
-  async.map(removeIds, doRemoveDatasetClientWithRecords, function(err, deleteResult) {
+  var datasetIds = _.uniq(_.pluck(datasetClientsToRemove, 'datasetId'));
+  debug('doRemoveDatasetClients: removed datasetClients = %d, datasets = %d', removeIds.length, datasetIds.length);
+  async.series([
+    function deleteDatasetClientAndRefs(callback) {
+      async.map(removeIds, doRemoveDatasetClientWithRecords, function(err, deleteResult) {
+        if (err) {
+          debug('Failed to delete datasetClients due to error %j', err);
```
david-martin: I think this might be better to console.error() this so the developer knows something isn't working right and they should take an action from it.
That action is to probably raise a bug against fh-mbaas-api

lib/sync/storage/dataset-clients.js
```diff
@@ -67,35 +86,47 @@ function doRemoveDatasetClients(datasetClientsToRemove, cb) {
...
+    function deleteDatasetClient(datasetClientJson, next) {
       var col = mongoClient.collection(DATASETCLIENTS_COLLECTION);
-      col.findOneAndDelete({'id': datasetClientId}, function(err, deleted) {
-        return callback(err, deleted);
-      });
+      col.findOneAndDelete({'id': datasetClientId}, next);
     }
   ], function(err, result) {
     if (err) {
       debug('Failed to doRemoveDatasetClientWithRecords due to error %j', err);
```
david-martin: Similar to above comment. Any error here should make it to logs even if the DEBUG flag is not set.

lib/sync/storage/dataset-clients.js
```diff
@@ -67,35 +86,47 @@ function doRemoveDatasetClients(datasetClientsToRemove, cb) {
...
+ * Remove unused records from the records collection for the given datasetId 
+ * @param {String} datasetId the datasetId 
+ * @param {Function} cb 
+ */
+function removeRecordsForDataset(datasetId, cb) {
+  debug('remove unused records from dataset %s', datasetId);
+  var recordsCollection = mongoClient.collection(getDatasetRecordsCollectionName(datasetId));
+  recordsCollection.deleteMany({'refs': {'$size': 0}}, function(err, result) {
+    if (err) {
+      debug('Failed to remove records from dataset %s due to error %j', datasetId, err);
```
david-martin: and one more

david-martin:
@wei-lee Reviewed. Above comments are all related to making it obvious to a user of this library that something when something has gone wrong (since debug statements are only output when DEBUG is set).

Taking a worst case scenario, dataset clients & records could be kept around forever because of an error. That error is supressed by default (debug()). The developer never sees that error.
Eventually, their collections sizes have grown massively and are impacting database and app performance.
This could at least be alerted to the user much earlier if these kinds of errors were output to stderr

wdyt?

lib/sync/storage/dataset-clients.js
wei-lee: yes, I think that makes sense.

wei-lee:
@david-martin I think I agree with your comments. But I think it might be better to review all the error statements and make changes in a separate PR to print out error messages. Perhaps in #132 ?

aidenkeating:
@wei-lee Yep, I'll make those changes in #132

PR #134, opened 2017-05-03
Prefix all metrics with a metricsKey

aidenkeating:
Currently any timing metric made by sync is not namespaced, meaning
that if two apps are running in the same environment the metrics
coming from them will not be able to be distinguished.

This change prefixes the timing metrics with the same prefix as the CPU
and memory metrics.


aidenkeating:
retest this please

aidenkeating:
@david-martin @wei-lee Would you mind taking a look?

wei-lee:
LGTM

PR #135, opened 2017-05-04
Fh 3561 map uids

wei-lee:
## Why

At the moment, if a record is being created on the client, and then updated, there is a chance that the update will not be applied because the update didn't get the server uid back to the client.

## What

The sync server will now save the record's client uid and server uid and replace the clientUid automatically if it's being used in the request.

ping @david-martin @aidenkeating @nialldonnellyfh  to review

### DIFF
--- a/integration/sync/test_uid_change.js
+++ b/integration/sync/test_uid_change.js
@@ -82,7 +82,7 @@ module.exports = {
           });
         },
         function wait(callback) {
-          setTimeout(callback, 80);
+          setTimeout(callback, 1000);
         },
         function checkDataUpdated(callback) {
           collection.findOne({'user': '1'}, function(err, found){
--- a/integration/sync/test_uid_change.js
+++ b/integration/sync/test_uid_change.js
@@ -1,0 +1,154 @@
+var sync = require('../../lib/sync');
+var assert = require('assert');
+var async = require('async');
+var helper = require('./helper');
+var util = require('util');
+
+var mongoDBUrl = 'mongodb://127.0.0.1:27017/test_sync_uid_change';
+var redisUrl = 'redis://127.0.0.1:6379';
+var DATASETID = 'syncUidChangeTest';
+var TESTCUID = 'syncUidChangeTestCuid';
+
+var mongodb;
+
+module.exports = {
+  'test sync data uid change': {
+    'before': function(done) {
+       sync.api.setConfig({
+        pendingWorkerInterval: 10, 
+        pendingWorkerBackoff: {strategy: 'none'},
+        ackWorkerInterval: 100, 
+        schedulerInterval: 100
+      });
+
+      async.series([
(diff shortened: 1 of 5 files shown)


integration/sync/test_uid_change.js
```diff
@@ -0,0 +1,154 @@
...
+      async.series([
+        function invokeSync(callback) {
+          sync.api.invoke(DATASETID, params, function(err, response){
+            assert.ok(!err, util.inspect(err));
+            assert.ok(response);
+            callback();
+          });
+        },
+        function wait(callback) {
+          setTimeout(callback, 100);
```
david-martin: @wei-lee why is a timeout needed here?

lib/sync/pending-processor.js
```diff
@@ -197,6 +201,20 @@ function doDelete(datasetId, pendingChange, callback) {
   });
 }
 
+function replaceUidIfRequired(datasetId, pendingChange, callback) {
+  syncStorage.lookupRecordUidByClientUid(datasetId, pendingChange.uid, function(err, serverUid){
+    if (err) {
+      debug('[%s] Failed to look up recordUid due to error %j', datasetId, err);
```
david-martin: @wei-lee In what cases would there be an err?
if the record wasn't found?

lib/sync/storage/dataset-clients.js
```diff
@@ -436,6 +443,52 @@ function doUpdateManyDatasetClients(query, fields, callback) {
...
+ * @param {String} datasetId the datasetId 
+ * @param {String} clientUid the client uid
+ * @param {String} serverUid the server uid
+ * @param {Function} callback
+ */
+function doSaveRecordUidMapping(datasetId, clientUid, serverUid, callback) {
+  debug('doSaveRecordUidMapping datasetId = %s, clientUid = %s, serverUid= %s', datasetId, clientUid, serverUid);
+  if (!clientUid || !serverUid) {
+    debug('invalid clientUid or serverUid. Nothing to save');
+    return callback();
```
david-martin: @wei-lee Is this an error scenario?
Is it possible for both of them to be not set?

integration/sync/test_uid_change.js
wei-lee: @david-martin we need to wait for all the pending changes applied before checking the database

lib/sync/pending-processor.js
wei-lee: a mongodb error more likely

wei-lee: if the record is not found, it won't return an error

lib/sync/storage/dataset-clients.js
wei-lee: I don't think so. If any of them is null, we just don't save them. It is possible for both of them to be null, but it is covered here?

wei-lee:
this PR is actually not required

PR #137, opened 2017-05-09
Add error event listener for Redis client

aidenkeating:
Currently, when Redis goes down or is not available the Redis client will
emit an `error` event. In node, if an 'error' event does not have a
listener node will throw an Error and stop the process.

This change adds a listener that will produce a `console.warn` log, this
will prevent the app from crashing when the Redis client is down.

### DIFF
--- a/lib/sync/index.js
+++ b/lib/sync/index.js
@@ -40,6 +40,8 @@ function connect(mongoDBConnectionUrl, mongoDBConnectionOption, redisUrl, cb) {
           url: redisUrl
         };
         var client = redis.createClient(redisOpts);
+        // We don't want the app to crash if Redis is not available.
+        client.on('error', handleRedisError);
         return callback(null, client);
       } else {
         return callback();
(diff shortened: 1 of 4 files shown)


lib/sync/index.js
```diff
@@ -57,6 +59,16 @@ function connect(mongoDBConnectionUrl, mongoDBConnectionOption, redisUrl, cb) {
...
 }
 
+/**
+ * Handle any errors emitted by the Redis client. Without a listener
+ * for the error event the process will exit when an event is emitted
+ *
+ * @param {Error} err An error produced by the Redis client
+ */
+function handleRedisError(err) {
+  console.warn('Error in Redis %j', err);
```
wei-lee: better use `%s` here as `%j` doesn't print out the error message

aidenkeating: Ooh, good to know

aidenkeating:
@wei-lee I'm not sure why that acceptance test is failing, especially since the same test (the test is duplicated for some reason) is passing on the same build and they're both passing locally.

wei-lee:
@aidenkeating I am not sure either. Perhaps just a timing issue. Will try it again

PR #138, opened 2017-05-11
Fh 3588 db pruning optimise

wei-lee:
a few improvements & bug fixes

* use mongodb TTL feature to manage the removal of queue data
* make sure only 1 datasetCleaner will remove data from mongodb when there are multiple workers
* fix an error when a sync operation is being performed on a datasetClient that is removed
* add `appname` to all the metrics data if it's available

### DIFF
--- a/lib/sync/datasetClientsCleaner.js
+++ b/lib/sync/datasetClientsCleaner.js
@@ -6,6 +6,7 @@ var syncUtil = require('./util');
 var debug = syncUtil.debug;
 var debugError = syncUtil.debugError;
 
+var syncLock;
 var storage;
 
 /**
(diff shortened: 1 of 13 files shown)


lib/sync/datasetClientsCleaner.js
```diff
@@ -22,26 +23,42 @@ var storage;
...
+        },
+        async.apply(storage.removeDatasetClients)
+      ], function(err, deletedDatasetClients) {
+        syncLock.release(lockName, lockCode, function(unlockErr){
+          if (unlockErr) {
+            debugError('Failed to release lock due to error %s', unlockErr);
+          }
+        });
+        if (err) {
+          debugError('Failed to cleanup dataset clients due to error %j', err);
```
aidenkeating: About the point you highlighted in a previous PR. Should this be %s as %j doesn't seem to include the error message?

wei-lee: yes, will change it

lib/sync/datasetClientsCleaner.js
```diff
@@ -22,26 +23,42 @@ var storage;
 function DatasetClientsCleaner(opts) {
   this.retentionPeriod = opts.retentionPeriod || '24h';
   this.checkFrequency = opts.checkFrequency || '1h';
+  this.cleanerLockName = opts.cleanerLockName || 'locks:sync:DatasetCleaner';
```
aidenkeating: Is `opts.cleanerLockName` able to be set by the user? It looks like it's not an option in sync-server and it not passed in?

wei-lee: it is, but TBH I don't think user will ever need to change it. Is there really value there to allow user to config it?

aidenkeating: Yeah, I think you're right here

lib/sync/datasetClientsCleaner.js
```diff
@@ -22,26 +23,42 @@ var storage;
...
+        },
+        async.apply(storage.removeDatasetClients)
+      ], function(err, deletedDatasetClients) {
+        syncLock.release(lockName, lockCode, function(unlockErr){
+          if (unlockErr) {
+            debugError('Failed to release lock due to error %s', unlockErr);
+          }
+        });
+        if (err) {
+          debugError('Failed to cleanup dataset clients due to error %s', err);
```
david-martin: Should this and above error use console.error? Should the developer need to know if the cleanup has failed for some reason?

lib/sync/mongodbQueue.js
```diff
@@ -34,10 +34,12 @@ function MongodbQueue(name, metrics, opts) {
   }
   this.queueName = name;
   this.metrics = metrics;
+  this.lock = lock;
+  this.lockName = opts.lockName || ('lock:sync:' + this.queueName);
+  this.lockTimeout = opts.lockTimeout || 10000;
```
david-martin: is 10 seconds enough for the task to finish when there are a lot of dataset clients?
I think there were numbers in excess of 10 seconds during load testing.

lib/sync/mongodbQueue.js
```diff
@@ -55,11 +57,75 @@ MongodbQueue.prototype.create = function(cb) {
...
+            return cb(err);
+          }
+          self.lock.release(self.lockName, lockCode, function(releaseErr){
+            if (releaseErr) {
+              debugError('[%s] failed to release lock due to error: %s', self.queueName, releaseErr);
+            }
+          });
+          return cb(null, self.queue);
+        });
+      } else {
```
david-martin: from a code readability point of view, it might be better to take case of this else case first as its kind of hidden at the bottom here

lib/sync/storage/dataset-clients.js
```diff
@@ -252,7 +252,9 @@ function doReadDatasetClientWithRecordsUseCache(datasetClientId, callback) {
         if (err) {
           return callback(err);
         }
-        cacheClient.set(cacheKey, JSON.stringify(results));
+        if (results) {
+          cacheClient.set(cacheKey, JSON.stringify(results));
```
david-martin: should an err be sent as first param if there are no results?
How does the caller handle a scenario where there is no error *and* no results object

david-martin:
@wei-lee reviewed with some comments. It looks like a good change overall.
We'll need to update/remove documentation since cleaning queues is automatic with this change.

lib/sync/datasetClientsCleaner.js
wei-lee: I believe `debugError` is actually calling to `console.error`

lib/sync/mongodbQueue.js
wei-lee: This lock is only for index creating on the queue collections. It really just to prevent multiple workers to create the indexes at the same time. I think 10 seconds should really be enough as it normally very quick to add indexes. It should have nothing to do with the number of dataset clients

lib/sync/mongodbQueue.js
wei-lee: ok, that makes sense

lib/sync/datasetClientsCleaner.js
david-martin: :+1:

lib/sync/mongodbQueue.js
david-martin: :+1:

lib/sync/mongodbQueue.js
david-martin: :+1:

lib/sync/storage/dataset-clients.js
wei-lee: the caller will have to handle that. It's like when you do db.find() - it could return null if it can't find the data.

The error should be only used if the request can't be completed.

david-martin: :+1:

PR #140, opened 2017-05-15
🐛 FH-3592 Ensure connect is finished before continuing to invoke sync

david-martin:
https://issues.jboss.org/browse/FH-3592

This change is to prevent a scenario where the application is
    starting and a client is sending a sync request to it.
    In that case, the `connect()` function may still be executing,
    so sync is not ready to `start` yet. To prevent `start` being
    executing in this case, an error will be sent to the sync client.
    This will cause a 'crashed' notification and the client will retry
    later.

### DIFF
--- a/lib/sync/index.js
+++ b/lib/sync/index.js
@@ -9,6 +9,7 @@ var async = require('async');
 var eventEmitter = new (require('events').EventEmitter)();
 var debug=syncUtil.debug;
 var debugError = syncUtil.debugError;
+var syncConnected = false;
 
 function toJSON(dataset_id, returnData, cb) {
   debug('[%s] toJSON',dataset_id);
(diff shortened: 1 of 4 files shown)


lib/sync/index.js
```diff
@@ -35,25 +36,22 @@ function connect(mongoDBConnectionUrl, mongoDBConnectionOption, redisUrl, cb) {
...
+        url: redisUrl
+      };
+      var client = redis.createClient(redisOpts);
+      // We don't want the app to crash if Redis is not available.
+      client.on('error', handleRedisError);
+      return callback(null, client);
     }
   ], function(err, results) {
-    if (err) return cb(err);
+    if (err) throw err;
```
wei-lee: just a small thing, if keep it consistently, let's add `{}` around the `throw` statement.

lib/sync/sync-server.js
```diff
@@ -233,7 +233,7 @@ function start(cb) {
       datasetClientCleaner.start(true, callback);
     }
   ], function(err) {
-    if (err) return cb(err);
+    if (err) throw err;
```
wei-lee: same thing here

lib/sync/index.js
wei-lee: also might worth adding a comment here to explain why we throw the error here rather than returning it in the callback

david-martin: :+1:

lib/sync/sync-server.js
david-martin: :+1:

lib/sync/index.js
```diff
@@ -90,7 +92,7 @@ function invoke(dataset_id, params, callback) {
 
   // Verify that fn param has been passed
   if (!params || !params.fn) {
-    var err = 'no fn parameter provided in params "' + util.inspect(params) + '"';
+    var err = new Error('no fn parameter provided in params "' + util.inspect(params) + '"');
     debugError('[%s] warn %s %j', dataset_id, err, params);
```
wojta: not sure, but err is no longer string, so debug won't probably show it properly, use %j instead

david-martin: Using %s results in `{}`, hence this PR https://github.com/feedhenry/fh-mbaas-api/pull/139. %j should only be used for plain old json objects.
e.g.
```javascript
var x = new Error('some message')
console.log(`(%s) (%j)`, x, x)
// (Error: some message) ({})
```

PR #142, opened 2017-05-15
Change the default sync worker interval & backoff settings

david-martin:
The default is to use an exponential backoff, ranging from 1ms to 1000ms.
    These values are based on load tests. The idea is to take full advantage of
    CPU when the sync queue is busy and to allow the CPU to be used for other
    things when the sync queue is not busy. Setting a max of 1000ms will ensure
    then should the server get busy again, there won't be a significant delay
    in ramping up.


david-martin:
@wei-lee If you could review the 3rd commit in the PR (The other 2 are in other PRs)
44d6db7

david-martin:
+ fh-mbaas-api@7.0.8

PR #143, opened 2017-05-16
RHMAP-15491 Fix background typo

aidenkeating:


aidenkeating:
@wei-lee @twnolan @jasonmadigan Would you mind taking a look?

aidenkeating:
@david-martin @wei-lee Why wasn't this caught in the tests?

jasonmadigan:
@aidenkeating change looks good, but I'd like to understand why our tests didn't catch too

david-martin:
@jasonmadigan @aidenkeating the integration tests did not catch this as they run against MongoDB 2.4.
This issue was discovered when using MongoDB 3.4 locally.
MongoDB 2.4 seems to ignore this typo.
Even if the tests ran against MongoDB 3.2 (same as whats in use on OpenShift), they would not have picked up on this issue.
This leads to a further question. If the 'background' typo meant the app didn't setup an index in the expected way, is it really needed? Should it be removed? Is there a dependency on it running in the background that a test scenario doesn't allow for?

Either way, extending the travis matrix to run against both mongo 2.4 & 3.2 (saas vs. openshift) should pick up on any potential issues that could happen for RHMAP users.
We can look at extending support to MongoDB 3.4 and adding that to the matrix too at the time.

david-martin:
@wei-lee @aidenkeating Adding mongo 3.2 to the build matrix in https://github.com/feedhenry/fh-mbaas-api/pull/144

PR #148, opened 2017-06-06
Increase timeout

sedroche:
Ticket: https://issues.jboss.org/browse/RHMAP-14109

# Motivation

As part of the above ticket, investigating the filtering of large datasets, the issue was that the filtering of large datasets could take a long time. This PR then intends to increase the timeout to match the timeout imposed by the AWS load balancer


camilamacedo86:
CLosing this PR, as discussed in the task related to this issue was defined that it is not the best approach.

PR #149, opened 2017-06-06
Update fh-sync dependency

thailekha:
Found this bug when working on https://issues.jboss.org/browse/RHMAP-15493

Version 1.0.1 should be used instead since there is a typo https://github.com/feedhenry/fh-sync/blob/482679ad1a60baac6d323e630f05439253ede2d3/lib/mongodbQueue.js#L119 in version 1.0.0 (https://github.com/feedhenry/fh-sync/blob/482679ad1a60baac6d323e630f05439253ede2d3/package.json#L3). The typo was fixed in version 1.0.1 https://github.com/feedhenry/fh-sync/blob/master/lib/mongodbQueue.js#L119

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,6 +1,6 @@
 {
   "name": "fh-mbaas-api",
-  "version": "7.0.9",
+  "version": "7.0.11",
   "dependencies": {
     "async": {
       "version": "2.1.5",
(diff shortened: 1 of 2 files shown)


npm-shrinkwrap.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fh-mbaas-api",
-  "version": "7.0.9",
+  "version": "7.0.10",
```
aidenkeating: It looks like `7.0.10` is already published, should this be `7.0.11` now as well as in `package.json`?

https://www.npmjs.com/package/fh-mbaas-api

thailekha: My bad. I Updated it.

wtrocki:
My bad.. I added ~ but also forgot we use use shrinkwrap for fh-mbaas-api :) 
Thanks for resolving this.

PR #156, opened 2017-07-14
downgrade fh-db from 2.1.0 to 2.0.0

rachael-oregan:
- downgrade fh-db to 2.0.0 because the query structure has changed in fh-db 2.1.0

- Jira: https://issues.jboss.org/browse/FH-3778

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,35 +1,35 @@
 {
   "name": "fh-mbaas-api",
-  "version": "7.0.13",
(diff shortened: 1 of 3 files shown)


david-martin:
@rachael-oregan :+1: you'll need to re-shrinkwrap as well. `├─┬ fh-db@2.1.0 invalid`

pb82:
@david-martin @rachael-oregan Should this also be a new version of fh-mbaas-api?

PR #171, opened 2017-10-26
Changed redirected link to point to actual target

brizrobbo:
@finp  Would you mind please reviewing when you get a chance.


brizrobbo:
@jasonmadigan Paul Wright (@finp) has approved this small docs change. Could you please merge when convenient? Thanks.

PR #172, opened 2017-11-02
Rhmap 18301

davidffrench:
JIRA: https://issues.jboss.org/browse/RHMAP-18301

When a request comes into an mbaas service, previously it used to only check the sent projectid against the list of authorised projects. This change makes a request to millicore to also check the sent app api key exists and is valid for one of the apps under the authorised projects.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "debug": "2.6.3",
     "fh-db": "3.3.0",
     "fh-mbaas-client": "0.15.0",
-    "fh-mbaas-express": "5.7.1",
+    "fh-mbaas-express": "~5.9.0",
     "fh-security": "0.2.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,14 +1,14 @@
...
+  "version": "8.1.0",
   "description": "FeedHenry MBAAS Cloud APIs",
   "main": "lib/api.js",
   "dependencies": {
     "async": "2.1.5",
     "debug": "2.6.3",
     "fh-db": "3.3.0",
     "fh-mbaas-client": "0.15.0",
-    "fh-mbaas-express": "5.7.1",
+    "fh-mbaas-express": "^5.9.0",
```
pb82: better use `~` to lock the version down to the patch level

PR #173, opened 2017-11-02
Update fh-mbaas-express for fh-mbaas-api version 6

davidffrench:
JIRA: https://issues.jboss.org/browse/RHMAP-18301

When a request comes into an mbaas service, previously it used to only check the sent projectid against the list of authorised projects. This change makes a request to millicore to also check the sent app api key exists and is valid for one of the apps under the authorised projects.

Updating for version 6 since the majority of customers are using version 6.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "eyes": "0.1.8",
     "fh-db": "1.4.4",
     "fh-mbaas-client": "0.16.5",
-    "fh-mbaas-express": "5.6.10",
+    "fh-mbaas-express": "~5.9.0",
     "fh-security": "0.2.0",
     "fh-statsc": "0.3.0",
     "lodash-contrib": "^393.0.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -10,7 +10,7 @@
     "eyes": "0.1.8",
     "fh-db": "1.4.4",
     "fh-mbaas-client": "0.16.5",
-    "fh-mbaas-express": "5.6.10",
+    "fh-mbaas-express": "^5.9.0",
```
pb82: better use `~` instead of  `^` to lock down to patch version

davidffrench:
@rachael-oregan there is a separate PR for the merge to master https://github.com/feedhenry/fh-mbaas-api/pull/172

package.json
davidffrench: I agree with this, making the change

PR #179, opened 2018-01-19
docs: adds note to include JIRA ref in commits

damienomurchu:
Update to README to include JIRA/ issue reference in commit messages. This small change echoes the one that is included in https://github.com/feedhenry/fh-fhc/pull/433 & is to maintain consistency


damienomurchu:
@camilamacedo86 a uniform/ default contribute.md sounds a very good idea to aim for

PR #181, opened 2018-02-27
Rhmap 19582

damienomurchu:
## Jira link
https://issues.jboss.org/browse/RHMAP-19582

### What
Update fh-sync dependency to use updated version of fh-sync

## Why
fh-sync needed to be updated so it did not rely on any github-hosted dependencies. fh-mbaas-api relied on fh-sync & now needs to be updated to use updated published version of fh-sync that doesn't rely on any github-hosted dependencies

## How
Update dependency to point to updated published version of fh-sync

## Verification steps
<to-be-added>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
-## [8.2.0] - Tue Feb 6
+## [8.2.1] - Tue Feb 27, 2018
+### Changed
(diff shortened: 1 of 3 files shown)


davidffrench:
@damienomurchu Please update the changelog for this change aswell

damienomurchu:
@davidffrench updated now

davidffrench:
@damienomurchu I would recommend giving as much detail as possible in the CHANGELOG. especially since you know the reason for the change. You could say something simple like updating fh-sync to remove a direct link to github in it's dependencies

damienomurchu:
@davidffrench good idea - updated now. Let me know if that note is sufficient (trying to walk the fine line between keeping the changelog informative but not noisy :) )

davidffrench:
@damienomurchu LGTM

PR #182, opened 2018-04-27
[task]:RHMAP-20281 -  Update CI process to use supported versions

camilamacedo86:
## JIRA
https://issues.jboss.org/browse/RHMAP-20279

## WHAT
Update NodeJS version to use 4 and 6 as follows.

```
node_js:
  - '4'
  - '6'
```

## WHY
Use in the CI the currently supported versions.

## Verification Steps:
Check if the all steps in Trevis are finishing with success.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,13 @@
 language: node_js
 sudo: required
 node_js:
-  - "4.4.3"
-  - "6.9.1"
+  - 4
+  - 6
 services:
   - docker
 before_install:
   - sudo apt-get update
   - sudo apt-get install --assume-yes apache2-utils
-  - npm install -g npm@2.13.5
   - npm install -g grunt-cli
   - npm config set strict-ssl false
 install: npm install
@@ -16,4 +15,4 @@ env:
   - MONGODB_VERSION=2.4
   - MONGODB_VERSION=3.2
 script:
-  - npm test
+  - grunt
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -3,7 +3,6 @@ var Memcached = require('memcached');
 var redisPort = 6379;
 var redisHost = '127.0.0.1';
 var redisPassword;
-var redisClient;
 var appname;
 var memcached;
 
--- a/lib/forms/index.js
+++ b/lib/forms/index.js
@@ -5,6 +5,10 @@ var assert = require('assert'),
   fs = require('fs'),
   exportSubmissions = require("./exportSubmissions"),
   config,
+  /* eslint-disable no-unused-vars */
+  appname,
+  fhutils,
+  // jshint ignore:end
   forms,
   exportCSVandPDF;
 


.travis.yml
```diff
@@ -1,14 +1,14 @@
...
-  - "6.9.1"
+  - '4'
+  - '6'
+  - '8'
 services:
   - docker
 before_install:
   - sudo apt-get update
   - sudo apt-get install --assume-yes apache2-utils
-  - npm install -g npm@2.13.5
```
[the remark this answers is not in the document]
camilamacedo86: When we install the NodeJS a default version of npm is installed and the old/deprecated versions received this upgrade to use this npm. However, I think you are right. Probably in the servers it still in this way, is better we keep it.

camilamacedo86:
@damienomurchu here for example. This lib is not building the NodeJS 8 as follows.

screen shot 2018-04-27 at 15 35 06

We need to fix it.

.travis.yml
```diff
@@ -1,14 +1,14 @@
 language: node_js
 sudo: required
 node_js:
-  - "4.4.3"
-  - "6.9.1"
+  - '4'
```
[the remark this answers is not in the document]
matskiv: fh-mbaas-api is used in cloud apps. So IMHO - as long as we offer option to deploy with Node 4, we should keep testing with Node 4.

camilamacedo86:
The PR is not working because it is not running working with the NodeJS 8.
The JIRA opened for we solved it is: https://issues.jboss.org/browse/RHMAP-20292

.travis.yml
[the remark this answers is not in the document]
camilamacedo86: @damienomurchu at the end was decided to remove this definition.

lib/cache.js
```diff
@@ -47,7 +47,7 @@ var jdgClient = function() {
   memcached = new Memcached(process.env.JDG_SERVICE_HOSTNAME + ':' + process.env.JDG_SERVICE_PORT, options);
 };
 
-var redisClient = function() {
+redisClient = function() {
```
grdryn: Would it be preferable here to instead remove the uninitialized `redisClient` from line 6, which would mean that the variable never has the undefined value? Visually this appears to be at the same scope level, and it's not (yet...thankfully) referenced before this.

lib/forms/index.js
```diff
@@ -453,19 +450,17 @@ function Forms() {
...
     });
   };
 }
 
-module.exports = function (cfg) {
+module.exports = function(cfg) {
   assert.ok(cfg, 'cfg is undefined');
   config = cfg;
-  appname = cfg.fhapi.appname;
-  fhutils = new futils(config);
```
grdryn: Can you confirm that there are no usages of these removed exports? Is it possible that a user could be depending on them (I haven't checked if they're exposed beyond this module)?

lib/push.js
```diff
@@ -38,14 +38,20 @@ module.exports = function (cfg) {
...
+    return function push(message, options, callback) {
+      if (!message) {
+        return callback(new Error("Missing required 'message' parameter"));
+      }
+      if (!options) {
+        return callback(new Error("Missing required 'options' parameter"));
+      }
       if (!options.broadcast) {
-        if (!options.apps) return callback(new Error("Missing required 'options.apps' parameter while 'options.broadcast' not specified" + JSON.stringify(options)));
+        if (!options.apps) {
```
grdryn: This is fine, but since we're changing it anyway, I think the two conditions can be merged. I think this will reduce the eslint warning count by 1 (no-lonely-if), and reduce the indentation level of this block.

I guess that would then be `if (!options.broadcast && !options.apps) {`

lib/cache.js
camilamacedo86: Done:-)

lib/push.js
camilamacedo86: Hi, in this case, it is not possible because the message for each is different. 

```
     if (!message) {
        return callback(new Error("Missing required 'message' parameter"));
      }
      if (!options) {
        return callback(new Error("Missing required 'options' parameter"));
      }

```

lib/forms/index.js
camilamacedo86: If you think that it has some reason then is better we keep it anyway.

camilamacedo86:
@grdryn  it is not correct. It is bronken with the master rebase. I will create a new one tks.

PR #184, opened 2018-05-01
Update CI process to use supported versions

camilamacedo86:
## JIRA
https://issues.jboss.org/browse/RHMAP-20296

## WHAT
* Update to use NodeJS version to use 4, 6, and 8 as follows.

```
node_js:
  - '4'
  - '6'
```
* Remove the fixed npm version
* Update to call the grunt tasks

## WHY
Use in the CI the currently supported versions.

## Verification Steps:
Check if the all steps in Trevis are finishing with success. 

PS.: This lib is not working with NodeJS 8 which will be checked in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,13 @@
 language: node_js
 sudo: required
 node_js:
-  - "4.4.3"
-  - "6.9.1"
+  - 4
+  - 6
 services:
   - docker
 before_install:
   - sudo apt-get update
   - sudo apt-get install --assume-yes apache2-utils
-  - npm install -g npm@2.13.5
   - npm install -g grunt-cli
   - npm config set strict-ssl false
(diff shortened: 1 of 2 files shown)


camilamacedo86:
@austincunningham in this case was defined not bump the version since the change is not in the app code. The change is just on the CI process which checks the PR's

davidffrench:
@camilamacedo86 I agree, i don't see a need for a version change. You can add in the ## [Unreleased] section to the top of the CHANGELOG as shown here - https://keepachangelog.com/en/1.0.0/.

camilamacedo86:
@davidffrench tks it is done.

camilamacedo86:
@austincunningham could you check if ok for you now?

camilamacedo86:
Thank you for the review.

PR #186, opened 2018-07-26
Rhmap 21000

davidffrench:
Bumping the version of fh-sync from 1.0.14 to 1.0.15 to resolve an issue where the message was not being returned to the client from the collision handler.


camilamacedo86:
Hi @davidffrench, I checked that the tag 1.0.5 exists then I am approving it. However, I did not check if it solves  "message was not being returned to the client from the collision handler"

davidffrench:
That is perfect @camilamacedo86 , thank you.

PR #188, opened 2018-08-22
 ⬆ upgrade fh deps to the latest version to get rid of error messages

wei-lee:
when the cloud app is running

To verify:

1. Clone the https://github.com/feedhenry-templates/helloworld-cloud app locally, and update the version of `fh-mbaas-api` to `9.0.0`.
2. Run `npm install --production` with node 10. It should finish ok. Start the app by running `node application.js`, you will see some error outputs.
3. Update the fh-mbaas-api version to `9.0.1-rc2` (same changes as this PR) and repeat the same steps, you should not see the error messages any more.


darahayes:
👀

PR #205, opened 2019-04-25
fix: reverts unifiedpush-node-sender dep to 0.12.1 to fix breaking change introduced (RHMAP-22117)

damienomurchu:
# Jira link(s)
https://issues.jboss.org/browse/RHMAP-22117

# What
Revert dependency bump

# Why
Fix breaking change introduced

# How
Revert dependency

# Verification Steps
<TODO>

## Checklist:

- [ ] Code has been tested locally by PR requester
- [ ] Changes have been successfully verified by another team member 

## Additional notes

Jenkinsfile removed for repo as part of this PR also. Reason for this is Jenkins instance has been switched off. TravisCI checks already in place, and as noted in PR the Jenkinsfile can be brought back if needed in the future for platform-specific needs.


damienomurchu:
@davidffrench fix as discussed. Have not checked CVE behind original bump, will do so & revert offline tomorrow before we look at merging this

grdryn:
@davidffrench I've triggered a build internally, to see if it will work there (not sure), that won't update the status check here though. If needs be, we can spin up the jenkins temporarily to build this. Alternatively, just remove the requirement for that status check, given that there's also a travis job. I wonder why there have been both up to now? :thinking:

davidffrench:
Thank you @grdryn and I do wonder why both are there as well. @damienomurchu will remove the Jenkinsfile now.

grdryn:
@davidffrench @damienomurchu it might be worth leaving the Jenkinsfile there for the moment in case it's run for platform release stuff? I don't think it is, but I'm not actually sure. Meh, we can always bring it back if necessary.

damienomurchu:
@grdryn will remove it if you're happy to see it removed. As you say we can bring it back if needed :)

grdryn:
Was just talking to @davidffrench briefly: I didn't realise that you were removing the Jenkinsfile to remove the required status check. That wouldn't have done that, so I've instead removed it from the branch protection configuration here in the GitHub repo settings.

damienomurchu:
thanks @grdryn - would you prefer I left the Jenkinsfile in the repo then?

PR #206, opened 2020-01-10
AGMS-1203: Rename package to community

wtrocki:
# Motivation 

Renaming package to `fh-mbaas-api-community` to hand over package to community with more fluent release schedule


darahayes:
I've approved but I really think final approval should be left to someone that's more familiar/involved with RHMAP.

wtrocki:
Agreed to just bump package instead

